PR #22, opened 2024-09-04
Trigger reconciliation of owner if change in controlled resource is detected

bruelea:
Before this change the reconciliation of a claim resource wasn't triggered if a change was detected in the owned resource because of the missing controller reference in the owner reference

Changes:
Use an owner reference with a reference to a controller
Remove status update which is no longer needed in ip address claim controller test
Remove unused functions from prefix claim type

### DIFF
--- a/internal/controller/ipaddressclaim_controller.go
+++ b/internal/controller/ipaddressclaim_controller.go
@@ -148,7 +148,7 @@ func (r *IpAddressClaimReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 		// 6.a create the IPAddress object
 		ipAddressResource := generateIpAddressFromIpAddressClaim(o, ipAddressModel.IpAddress)
-		err = controllerutil.SetOwnerReference(o, ipAddressResource, r.Scheme)
+		err = controllerutil.SetControllerReference(o, ipAddressResource, r.Scheme)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
--- a/api/v1/prefixclaim_types.go
+++ b/api/v1/prefixclaim_types.go
@@ -17,9 +17,6 @@ limitations under the License.
 package v1
 
 import (
-	"strings"
-
-	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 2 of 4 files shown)


internal/controller/ipaddressclaim_controller.go
```diff
@@ -148,7 +148,7 @@ func (r *IpAddressClaimReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 		// 6.a create the IPAddress object
 		ipAddressResource := generateIpAddressFromIpAddressClaim(o, ipAddressModel.IpAddress)
-		err = controllerutil.SetOwnerReference(o, ipAddressResource, r.Scheme)
+		err = controllerutil.SetControllerReference(o, ipAddressResource, r.Scheme)
```
henrybear327, line 151: Can you help me understand what's the rationale behind this change :)

bruelea: By default the `Owns()` function in the `SetupWithManager` function reconciles the first controller-type owner reference. With this change a controller-type owner reference is added to the ipAddressResource and `Owns()` will trigger the reconciliation of the ip address claim resource if a change in the ip address resource is detected.

henrybear327 (approved):

henrybear327:
@bruelea I think there are 2 extra `SetOwnerReference`s left there
-  in https://github.com/netbox-community/netbox-operator/blob/798a3082fc8b6703614da52b613ef89933a8f649/internal/controller/ipaddress_controller_test.go#L267 that needs to be changed. 
- in https://github.com/netbox-community/netbox-operator/blob/798a3082fc8b6703614da52b613ef89933a8f649/api/v1/prefixclaim_types.go#L126. It's an unused function though, maybe we can just remove the function?

bruelea:
> @bruelea I think there are 2 extra `SetOwnerReference`s left there
> 
> * in https://github.com/netbox-community/netbox-operator/blob/798a3082fc8b6703614da52b613ef89933a8f649/internal/controller/ipaddress_controller_test.go#L267
>    that needs to be changed.
> * in https://github.com/netbox-community/netbox-operator/blob/798a3082fc8b6703614da52b613ef89933a8f649/api/v1/prefixclaim_types.go#L126
>   . It's an unused function though, maybe we can just remove the function?

Good catch! I removed the unused functions and replaced the function in the controller test.

henrybear327 (approved):

PR #30, opened 2024-09-04
Add image build pipeline

henrybear327:
- Build docker images for both x86 and ARM architectures
- Perform image build test on PRs
- Push to ghcr.io when merged to `main`

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -1,0 +1,54 @@
+---
+name: Build docker image
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - v*
+  pull_request:
+env:
+  IMAGE_NAME: netbox-operator
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


henrybear327:
Only the images of those commits that are merged into the main branch
will be available on the GitHub registry, as we don't have unlimited
storage up here [1].

The pipeline is tested at https://github.com/henrybear327/netbox-operator/pull/1.

Reference:
- https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages

jstudler:
> Building an ARM docker image once will take around 17 minutes per run. See https://github.com/netbox-community/netbox-operator/actions/runs/10770458614/job/29863858976?pr=30
> 
> But it's a nice thing to have as we can finally run the operator on M1 natively!

Wonder if we really want to build on MRs. But we can always remove it later in case we use too much CI resources.

jstudler (approved):
lgtm

alexandernorth (approved):
lgtm

PR #43, opened 2024-09-06
Restructure the controller tests

bruelea:
restructure controller tests to avoid race conditions and make it compatible with sigs.k8s.io/controller-runtime v0.19.0


jstudler (approved):
looks good to me

PR #60, opened 2024-09-13
Add license header lint rules

henrybear327:
Making sure that the header is properly added


henrybear327:
@jstudler I think we can merge this for now, as our conclusion regarding license header compliance basically dictates that we need to have an Apache license on top of all source files in all source files.

WDYT?

alexandernorth (approved):

PR #72, opened 2024-09-23
Add check for tenant existence

### LINKED ISSUE #39: IP assigned when owner does not exist in Netbox
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via netbox operator maintainers.
- [X] Existing open issues have been checked and this is not a duplicate.

### What happened?

When creating an IpAddressClaim with a tenant which does not exist, an ipAddress resource is generated containing the next available IPAddress from the prefix.

### What did you expect to happen?

There should be no IPAddress generated at all since the Tenant does not exist.

### How can we reproduce it (as minimally and precisely as possible)?

Apply following cr and an ipaddress resource will be generated containing an ip:
```yaml
apiVersion: netbox.dev/v1
kind: IpAddressClaim
metadata:
  labels:
    app.kubernetes.io/name: netbox-operator
    app.kubernetes.io/managed-by: kustomize
  name: ipaddressclaim-sample-2
spec:
  tenant: "non-existant-tenant"
  description: "some description"
  comments: "your comments"
  preserveInNetbox: true
  parentPrefix: "2.0.0.0/16"
```
### Netbox operator version

Latest from main branch

### Netbox operator configuration (command line flags or environment variables)

Followed the instructions from the readme

### Relevant log output

```Shell
IP Addres generated from claim above:

Name:         ipaddressclaim-sample-2
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  netbox.dev/v1
Kind:         IpAddress
Metadata:
  Creation Timestamp:  2024-09-05T13:16:10Z
  Generation:          1
  Owner References:
    API Version:     netbox.dev/v1
    Kind:            IpAddressClaim
    Name:            ipaddressclaim-sample-2
    UID:             c568f9e5-1964-4f31-aab3-170cac459dea
  Resource Version:  5641
  UID:               493a998c-3753-4962-9ee1-932738ad0b91
Spec:
  Comments:  your comments
  Custom Fields:
    Netbox Operator Restoration Hash:  0e12f7d8e25476025adf6e5dba63c8b60f674c80
  Description:                         some description
  Ip Address:                          2.0.0.2/32
  Preserve In Netbox:                  true
  Tenant:                              non-existant-tenant
Status:
  Conditions:
    Last Transition Time:  2024-09-05T13:16:10Z
    Message:               Failed to reserve IP in NetBox. 2.0.0.2/32
    Reason:                FailedToReserveIpInNetbox
    Status:                False
    Type:                  Ready
Events:
  Type     Reason                     Age   From                   Message
  ----     ------                     ----  ----                   -------
  Warning  FailedToReserveIpInNetbox  4m    ip-address-controller  Failed to reserve IP in NetBox. 2.0.0.2/32
```

### Anything else we need to know?

_No response_

MaIT-HgA:
Close #39


bruelea (changes requested):
The check is only required if a new ip address is assigned. Instead of directly checking this in the reconcile loop we could check it in the GetAvailableIpAddressClaim (https://github.com/netbox-community/netbox-operator/blob/508165e68b2db3275e70f43140f736df23a3f217/pkg/netbox/api/ip_address_claim.go#L59) function.

pkg/netbox/api/ip_address_claim.go
faebr, line 1: Could we also add a new testcase which covers this fix: claim when tenant not in netbox?

internal/controller/ipaddressclaim_controller_test.go
```diff
@@ -201,7 +201,9 @@ var _ = Describe("IpAddressClaim Controller", Ordered, func() {
 			defaultIpAddressClaimCR(), defaultIpAddressCreatedByClaim(false), nil,
 			nil,
 			nil,
-			nil,
+			[]func(*mock_interfaces.MockTenancyInterface, chan error){
```
bruelea, line 204: this change is not required. If the existence of the tenant is checked in the GetAvailableIpAddressByClaim function, the TenancyTenantsList endpoint is not required for this test case.

faebr:
I just did a manual check, in my opinion the operator should not log the stack trace. This makes it look like a reconciler error. Its ok when the loglevel is error but adding the stack trace is not needed.
Example:
`2024-09-26T13:48:31+02:00	INFO	reconcile loop started	{"controller": "ipaddressclaim", "controllerGroup": "netbox.dev", "controllerKind": "IpAddressClaim", "IpAddressClaim": {"name":"ipaddressclaim-sample-2","namespace":"default"}, "namespace": "default", "name": "ipaddressclaim-sample-2", "reconcileID": "6d11f8c1-78bf-44ca-a2b7-01b42f3d7459"}
2024-09-26T13:48:31+02:00	ERROR	Reconciler error	{"controller": "ipaddressclaim", "controllerGroup": "netbox.dev", "controllerKind": "IpAddressClaim", "IpAddressClaim": {"name":"ipaddressclaim-sample-2","namespace":"default"}, "namespace": "default", "name": "ipaddressclaim-sample-2", "reconcileID": "6d11f8c1-78bf-44ca-a2b7-01b42f3d7459", "error": "tenant not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/Users/taascfas/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/Users/taascfas/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:263
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
	/Users/taascfas/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:224`

Also when a claim succeeds a status condition is set, would it make sense to also set a status condition when tenant is not found?

I see that this is the same when the parent-prefix does not exist, could we also change this? @bruelea @alexandernorth @jstudler 
`2024-09-26T13:54:05+02:00	ERROR	Reconciler error	{"controller": "ipaddressclaim", "controllerGroup": "netbox.dev", "controllerKind": "IpAddressClaim", "IpAddressClaim": {"name":"ipaddressclaim-sample","namespace":"default"}, "namespace": "default", "name": "ipaddressclaim-sample", "reconcileID": "7204d041-cc55-40d4-80d1-bc252fcbcdd3", "error": "parent prefix not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/Users/taascfas/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/Users/taascfas/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:263
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
	/Users/taascfas/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:224`

bruelea:
> Also when a claim succeeds a status condition is set, would it make sense to also set a status condition when tenant is not found?

Agree, the status condition should be updated too. In https://github.com/netbox-community/netbox-operator/pull/74 this is implemented.

MaIT-HgA:
@bruelea Should I let you update the condition since you've already implemented it?

@faebr I'll take a look into k8s events for error logs

jstudler:
Maybe I'm getting it wrong but wouldn't the simplest solution be to just add a function like e.g. "TenantExists" which takes a tenant string and returns a bool and then check this in e.g. https://github.com/netbox-community/netbox-operator/blob/main/internal/controller/ipaddressclaim_controller.go#L79 and https://github.com/netbox-community/netbox-operator/blob/main/internal/controller/ipaddress_controller.go#L103 (and the respective location in the prefix controllers)? It's kind of a config validation and should happen early in the process. If it doesn't exist, we write a log entry and exit. 

With the proposed solution it will create a lease and add other complexity that we don't need. wdyt @MaIT-HgA @faebr @bruelea ?

internal/controller/ipaddressclaim_controller.go
```diff
@@ -137,9 +137,17 @@ func (r *IpAddressClaimReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 				})
 			if err != nil {
-				return ctrl.Result{}, err
+				setConditionErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpAssignedFalse, corev1.EventTypeWarning, err.Error())
+				if setConditionErr != nil {
+					return ctrl.Result{}, fmt.Errorf("error updating status: %w, when getting an available IP address failed: %w", setConditionErr, err)
+				}
+
+				// since it's a Netbox error and not reconcile error, the log is of type Info and not Error
+				debugLogger.Info(fmt.Sprintf("error in getting available IP address: %s", err.Error()))
```
bruelea, line 146: The event will already create a log of the error. We could skip this log with the dubugLogger.

internal/controller/prefixclaim_controller.go
faebr, line 1: should this be here and not in pr 74 as commented by @MaIT-HgA ?

the status condition and the events will be implemented in https://github.com/netbox-community/netbox-operator/pull/74

bruelea:
> Maybe I'm getting it wrong but wouldn't the simplest solution be to just add a function like e.g. "TenantExists" which takes a tenant string and returns a bool and then check this in e.g. https://github.com/netbox-community/netbox-operator/blob/main/internal/controller/ipaddressclaim_controller.go#L79 and https://github.com/netbox-community/netbox-operator/blob/main/internal/controller/ipaddress_controller.go#L103 (and the respective location in the prefix controllers)? It's kind of a config validation and should happen early in the process. If it doesn't exist, we write a log entry and exit.
> 
> With the proposed solution it will create a lease and add other complexity that we don't need. wdyt @MaIT-HgA @faebr @bruelea ?

That would work too. The motivation to do the check if the tenant exists was that the check is only required if the IP is not yet assigned (https://github.com/netbox-community/netbox-operator/pull/72#pullrequestreview-2325669910). We could check the "IPAssigned" status condition in the reconcile loop and check if the tenant exists if the status condition is "false".

jstudler:
> > Maybe I'm getting it wrong but wouldn't the simplest solution be to just add a function like e.g. "TenantExists" which takes a tenant string and returns a bool and then check this in e.g. https://github.com/netbox-community/netbox-operator/blob/main/internal/controller/ipaddressclaim_controller.go#L79 and https://github.com/netbox-community/netbox-operator/blob/main/internal/controller/ipaddress_controller.go#L103 (and the respective location in the prefix controllers)? It's kind of a config validation and should happen early in the process. If it doesn't exist, we write a log entry and exit.
> > With the proposed solution it will create a lease and add other complexity that we don't need. wdyt @MaIT-HgA @faebr @bruelea ?
> 
> That would work too. The motivation to do the check if the tenant exists was that the check is only required if the IP is not yet assigned ([#72 (review)](https://github.com/netbox-community/netbox-operator/pull/72#pullrequestreview-2325669910)). We could check the "IPAssigned" status condition in the reconcile loop and check if the tenant exists if the status condition is "false".

True, this is a valid concern for the (rare?) case that the tenant is removed from NetBox after the ip/ipc/px/pxc resource(s) have been sucessfully created. So maybe we keep it as is.

internal/controller/prefixclaim_controller.go
MaIT-HgA: We discussed that I would add an event for the tenancy check and merge this MR, @bruelea will then rebase from main

internal/controller/ipaddress_controller.go
```diff
@@ -101,6 +101,15 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	}
 
+	// if PreserveIpInNetbox flag is false then register finalizer if not yet registered
```
bruelea, line 104: Currently the finalizer won't be removed if the deletion of the ipaddress in NetBox fails. This also includes the case where the resource doesn't exist.

Adding the finalizer before the resource was created in NetBox could lead to resources that cannot be deleted by the operator. Should we edit the reconcile loop to only try the deletion if Status.IpAddressId is not `0` and to not found errors from `NetboxClient.DeleteIpAddress`.

MaIT-HgA:
Finalizer changes is removed from this MR and will be in another MR

bruelea (approved):
Looks good!

PR #74, opened 2024-09-24
Add support for IPv6

bruelea:
changes:
- edit validation rule for ip addresses and prefixes in the CRD definition to allow IPv6
- in the controllers the leaselock name was replaced by the parentprefix in the logs about leaselocker
- add the update the status conditions if the assignment of a prefix/ipaddress fails in the claim controllers
- make the "github.com/netbox-community/netbox-operator/pkg/netbox/api" package compatible with IPv6 and update the tests
- remove unused function

### DIFF
--- a/pkg/netbox/api/prefix_claim_test.go
+++ b/pkg/netbox/api/prefix_claim_test.go
@@ -226,6 +226,22 @@ func TestPrefixClaim_InvalidIPv4PrefixLength(t *testing.T) {
 	mockPrefixIpam := mock_interfaces.NewMockIpamInterface(ctrl)
 	mockTenancy := mock_interfaces.NewMockTenancyInterface(ctrl)
 
+	// example of tenant
+	tenantId := int64(2)
+	tenantName := "Tenant1"
+	tenantOutputSlug := "tenant1"
+	expectedTenant := &tenancy.TenancyTenantsListOK{
+		Payload: &tenancy.TenancyTenantsListOKBody{
+			Results: []*netboxModels.Tenant{
+				{
+					ID:   tenantId,
+					Name: &tenantName,
+					Slug: &tenantOutputSlug,
+				},
+			},
+		},
+	}
+
 	parentPrefix := "10.112.140.0/24"
 	parentPrefixId := int64(1)
 	prefixListInput := ipam.
@@ -246,7 +262,10 @@ func TestPrefixClaim_InvalidIPv4PrefixLength(t *testing.T) {
 		},
 	}
 
+	inputTenant := tenancy.NewTenancyTenantsListParams().WithName(&tenantName)
+
 	mockPrefixIpam.EXPECT().IpamPrefixesList(prefixListInput, nil).Return(prefixListOutput, nil).AnyTimes()
+	mockTenancy.EXPECT().TenancyTenantsList(inputTenant, nil).Return(expectedTenant, nil).AnyTimes()
 
 	netboxClient := &NetboxClient{
 		Ipam:    mockPrefixIpam,
@@ -256,6 +275,9 @@ func TestPrefixClaim_InvalidIPv4PrefixLength(t *testing.T) {
 	actual, err := netboxClient.GetAvailablePrefixByClaim(&models.PrefixClaim{
 		ParentPrefix: parentPrefix,
 		PrefixLength: "/33",
+		Metadata: &models.NetboxMetadata{
+			Tenant: tenantName,
+		},
 	})
 
 	var expectedPrefix *models.Prefix
@@ -270,6 +292,22 @@ func TestPrefixClaim_FailWhenRequestingEntirePrefix(t *testing.T) {
 	mockPrefixIpam := mock_interfaces.NewMockIpamInterface(ctrl)
 	mockTenancy := mock_interfaces.NewMockTenancyInterface(ctrl)
 
+	// example of tenant
+	tenantId := int64(2)
+	tenantName := "Tenant1"
+	tenantOutputSlug := "tenant1"
+	expectedTenant := &tenancy.TenancyTenantsListOK{
+		Payload: &tenancy.TenancyTenantsListOKBody{
+			Results: []*netboxModels.Tenant{
+				{
+					ID:   tenantId,
+					Name: &tenantName,
+					Slug: &tenantOutputSlug,
+				},
+			},
+		},
+	}
+
 	parentPrefix := "10.112.140.0/24"
 	parentPrefixId := int64(1)
 	prefixListInput := ipam.
@@ -290,7 +328,10 @@ func TestPrefixClaim_FailWhenRequestingEntirePrefix(t *testing.T) {
 		},
 	}
 
+	inputTenant := tenancy.NewTenancyTenantsListParams().WithName(&tenantName)
+
 	mockPrefixIpam.EXPECT().IpamPrefixesList(prefixListInput, nil).Return(prefixListOutput, nil).AnyTimes()
+	mockTenancy.EXPECT().TenancyTenantsList(inputTenant, nil).Return(expectedTenant, nil).AnyTimes()
 
 	netboxClient := &NetboxClient{
 		Ipam:    mockPrefixIpam,
@@ -300,6 +341,9 @@ func TestPrefixClaim_FailWhenRequestingEntirePrefix(t *testing.T) {
 	actual, err := netboxClient.GetAvailablePrefixByClaim(&models.PrefixClaim{
 		ParentPrefix: parentPrefix,
 		PrefixLength: "/24",
+		Metadata: &models.NetboxMetadata{
+			Tenant: tenantName,
+		},
 	})
 
 	var expectedPrefix *models.Prefix
@@ -314,6 +358,22 @@ func TestPrefixClaim_FailWhenPrefixLargerThanParent(t *testing.T) {
 	mockPrefixIpam := mock_interfaces.NewMockIpamInterface(ctrl)
 	mockTenancy := mock_interfaces.NewMockTenancyInterface(ctrl)
 
+	// example of tenant
+	tenantId := int64(2)
+	tenantName := "Tenant1"
+	tenantOutputSlug := "tenant1"
+	expectedTenant := &tenancy.TenancyTenantsListOK{
+		Payload: &tenancy.TenancyTenantsListOKBody{
+			Results: []*netboxModels.Tenant{
+				{
+					ID:   tenantId,
+					Name: &tenantName,
+					Slug: &tenantOutputSlug,
+				},
+			},
+		},
+	}
+
 	parentPrefix := "10.112.140.0/24"
 	parentPrefixId := int64(1)
 	prefixListInput := ipam.
@@ -334,7 +394,10 @@ func TestPrefixClaim_FailWhenPrefixLargerThanParent(t *testing.T) {
 		},
 	}
 
+	inputTenant := tenancy.NewTenancyTenantsListParams().WithName(&tenantName)
+
 	mockPrefixIpam.EXPECT().IpamPrefixesList(prefixListInput, nil).Return(prefixListOutput, nil).AnyTimes()
+	mockTenancy.EXPECT().TenancyTenantsList(inputTenant, nil).Return(expectedTenant, nil).AnyTimes()
 
 	netboxClient := &NetboxClient{
 		Ipam:    mockPrefixIpam,
@@ -344,6 +407,9 @@ func TestPrefixClaim_FailWhenPrefixLargerThanParent(t *testing.T) {
 	actual, err := netboxClient.GetAvailablePrefixByClaim(&models.PrefixClaim{
 		ParentPrefix: parentPrefix,
 		PrefixLength: "/20",
+		Metadata: &models.NetboxMetadata{
+			Tenant: tenantName,
+		},
 	})
 
 	var expectedPrefix *models.Prefix
@@ -418,6 +484,9 @@ func TestPrefixClaim_ValidIPv6PrefixLength(t *testing.T) {
 	actual, err := netboxClient.GetAvailablePrefixByClaim(&models.PrefixClaim{
 		ParentPrefix: parentPrefix,
 		PrefixLength: "/33",
+		Metadata: &models.NetboxMetadata{
+			Tenant: tenantName,
+		},
 	})
 
 	assert.Nil(t, err)


api/v1/prefixclaim_types.go
```diff
@@ -29,12 +29,12 @@ type PrefixClaimSpec struct {
...
 
 	//+kubebuilder:validation:Required
-	//+kubebuilder:validation:Pattern=`^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))\/([1-9]|[12][0-9]|3[0-2])$`
+	//+kubebuilder:validation:Format=cidr
 	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'parentPrefix' is immutable"
 	ParentPrefix string `json:"parentPrefix"`
 
 	//+kubebuilder:validation:Required
-	//+kubebuilder:validation:Pattern=`^\/([1-9]|[12][0-9]|3[0-2])$`
+	//+kubebuilder:validation:Pattern=`^\/[0-9]|[1-9][0-9]|11[0-9]|12[0-8]$`
```
henrybear327, line 37: I think this will reject the range (inclusive) `100` to `109`, right?

config/crd/bases/netbox.dev_prefixclaims.yaml
```diff
@@ -60,13 +60,13 @@ spec:
...
               parentPrefix:
-                pattern: ^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))\/([1-9]|[12][0-9]|3[0-2])$
+                format: cidr
                 type: string
                 x-kubernetes-validations:
                 - message: Field 'parentPrefix' is immutable
                   rule: self == oldSelf
               prefixLength:
-                pattern: ^\/([1-9]|[12][0-9]|3[0-2])$
+                pattern: ^\/[0-9]|[1-9][0-9]|11[0-9]|12[0-8]$
```
henrybear327, line 69: ditto

henrybear327 (changes requested):

api/v1/prefixclaim_types.go
bruelea: Thanks! I fixed the pattern to also allow prefix lengths 100-109

internal/controller/prefixclaim_controller.go
```diff
@@ -105,18 +106,22 @@ func (r *PrefixClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 		// 4. try to reclaim Prefix using restorationHash
 		h := generatePrefixRestorationHash(prefixClaim)
 		prefixModel, err := r.NetboxClient.RestoreExistingPrefixByHash(h)
 		if err != nil {
+			if setConditionErr := r.SetConditionAndCreateEvent(ctx, prefixClaim, netboxv1.ConditionPrefixAssignedFalse, corev1.EventTypeWarning, err.Error()); setConditionErr != nil {
+				return ctrl.Result{}, fmt.Errorf("error updating status: %w, when look up of prefix by hash failed: %w", setConditionErr, err)
+			}
 			return ctrl.Result{}, err
 		}
 		// TODO: set condition for each error
```
henrybear327, line 127: nit: remove this TODO :)

pkg/netbox/api/ip_address_claim_test.go
```diff
@@ -35,20 +35,33 @@ func TestIPAddressClaim(t *testing.T) {
...
+				Address: ipAddressV4_2,
+				Family:  int64(4),
+			},
 		}
 	}
 
+	// test data for IPv6 ip address claim
+	parentPrefixIdV6 := int64(4)
+	parentPrefixV6 := "2001:db8:85a3:8d3::/64"
+	ipAddressV6 := "2001:db8:85a3:8d3::2/64"
```
henrybear327, line 62: Question: is there a specific reason we don't do `ipAddressV6_1` and `ipAddressV6_2`, like we have for IPv4? :)

pkg/netbox/api/prefix_claim.go
```diff
@@ -53,16 +53,31 @@ func (r *NetboxClient) RestoreExistingPrefixByHash(hash string) (*models.Prefix,
 	}, nil
 }
 
-func isRequestingTheEntireParentPrefix(prefixClaim *models.PrefixClaim) (bool, error) {
+func isRequestingInvalidPrefixLength(prefixClaim *models.PrefixClaim, prefixFamily int64) (bool, string, error) {
```
henrybear327, line 56: Is there a specific reason to do `(bool, string, error)` instead of just `(err)`?

As the only use case is 
```
if ret, message, err := isRequestingInvalidPrefixLength(prefixClaim,
	*responseParentPrefix.Payload.Results[0].Family.Value); err != nil {
	return nil, err
} else if ret {
	return nil, errors.New("invalid prefix length: " + message)
}
```
I am not sure if splitting out `ret` and then generating an error from the string using `errors.New` makes sense. :)

pkg/netbox/api/prefix_claim.go
```diff
@@ -53,16 +53,31 @@ func (r *NetboxClient) RestoreExistingPrefixByHash(hash string) (*models.Prefix,
...
+	if err != nil {
+		return false, "", err
+	}
+
+	if parentPrefixLength == requestedPrefixLength {
+		return true, "requesting the entire parent prefix range is disallowed", nil
+	} else if parentPrefixLength > requestedPrefixLength {
+		return true, "requested prefix size must be smaller than the parent prefix size", nil
+	} else if prefixFamily == int64(IPv4Familiy) && requestedPrefixLength > 32 {
+		return true, "requested prefix length must be smaller than 32 for IPv4", nil
```
henrybear327, line 77: Since we are on it, I think we can also add `requested prefix length must be smaller than 128 for IPv6`! :)

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -43,8 +43,14 @@ func TestPrefixClaim_GetAvailablePrefixesByParentPrefix(t *testing.T) {
 	childPrefix2 := "10.120.180.0/24"
 	availablePrefixListOutput := &ipam.IpamPrefixesAvailablePrefixesListOK{
 		Payload: []*netboxModels.AvailablePrefix{
-			{Prefix: childPrefix1},
-			{Prefix: childPrefix2},
+			{
+				Prefix: childPrefix1,
+				Family: 4,
```
henrybear327, line 48: nit: use `int64(IPv4Familiy)`

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -43,8 +43,14 @@ func TestPrefixClaim_GetAvailablePrefixesByParentPrefix(t *testing.T) {
...
+			},
+			{
+				Prefix: childPrefix2,
+				Family: 4,
```
henrybear327, line 52: ditto

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -202,6 +213,210 @@ func TestPrefixClaim_GetBestFitPrefixByClaim(t *testing.T) {
...
+	mockPrefixIpam := mock_interfaces.NewMockIpamInterface(ctrl)
+	mockTenancy := mock_interfaces.NewMockTenancyInterface(ctrl)
+
+	parentPrefix := "10.112.140.0/24"
+	parentPrefixId := int64(1)
+	prefixListInput := ipam.
+		NewIpamPrefixesListParams().
+		WithPrefix(&parentPrefix)
+
+	prefixFamily := int64(4)
```
henrybear327, line 228: ditto

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -202,6 +213,210 @@ func TestPrefixClaim_GetBestFitPrefixByClaim(t *testing.T) {
...
```
henrybear327, line 272: ditto

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -202,6 +213,210 @@ func TestPrefixClaim_GetBestFitPrefixByClaim(t *testing.T) {
...
```
henrybear327, line 316: ditto

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -202,6 +213,210 @@ func TestPrefixClaim_GetBestFitPrefixByClaim(t *testing.T) {
...
+	}
+
+	parentPrefix := "2001:db8:85a3:8d3::/30"
+	parentPrefixId := int64(1)
+	prefix := "2001:db8:85a3:8d3::/33"
+	prefixListInput := ipam.
+		NewIpamPrefixesListParams().
+		WithPrefix(&parentPrefix)
+
+	prefixFamily := int64(6)
```
henrybear327, line 377: ditto

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -232,12 +447,16 @@ func TestPrefixClaim_GetBestFitPrefixByClaimNoAvailablePrefixMatchesSize(t *test
 	prefixListInput := ipam.
 		NewIpamPrefixesListParams().
 		WithPrefix(&parentPrefix)
+
+	prefixFamily := int64(4)
```
henrybear327, line 451: ditto

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -306,12 +525,16 @@ func TestPrefixClaim_GetBestFitPrefixByClaimNoAvailablePrefixMatchesSizeCriteria
...
```
henrybear327, line 529: ditto

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -375,12 +598,16 @@ func TestPrefixClaim_GetBestFitPrefixByClaimInvalidFormatFromNetbox(t *testing.T
...
```
henrybear327, line 602: ditto

pkg/netbox/api/prefix_claim_test.go
```diff
@@ -449,12 +676,16 @@ func TestPrefixClaim_GetBestFitPrefixByClaimInvalidPrefixClaim(t *testing.T) {
...
```
henrybear327, line 680: ditto

pkg/netbox/api/ip_address_claim_test.go
```diff
@@ -35,20 +35,33 @@ func TestIPAddressClaim(t *testing.T) {
...
+	singleIpAddressV4_2 := "10.112.140.2/32"
+
+	// example of available IPv4 IP addresses
+	availAddressesIPv4 := func() []*netboxModels.AvailableIP {
 		return []*netboxModels.AvailableIP{
-			{Address: ipAddress1},
-			{Address: ipAddress2},
+			{
+				Address: ipAddressV4_1,
+				Family:  int64(4),
```
henrybear327, line 50: nit: Use `int64(IPv4Familiy)`

pkg/netbox/api/ip_address_claim_test.go
```diff
@@ -35,20 +35,33 @@ func TestIPAddressClaim(t *testing.T) {
...
+			},
+			{
+				Address: ipAddressV4_2,
+				Family:  int64(4),
```
henrybear327, line 54: ditto

pkg/netbox/api/ip_address_claim_test.go
```diff
@@ -81,37 +94,40 @@ func TestIPAddressClaim(t *testing.T) {
...
 		}
 
-		inputIps := ipam.NewIpamPrefixesAvailableIpsListParams().WithID(parentPrefixId)
+		inputIps := ipam.NewIpamPrefixesAvailableIpsListParams().WithID(parentPrefixIdV4)
 		outputIps := &ipam.IpamPrefixesAvailableIpsListOK{
 			Payload: []*netboxModels.AvailableIP{
-				{Address: ipAddress2},
+				{
+					Address: ipAddressV4_2,
+					Family:  int64(4),
```
henrybear327, line 129: ditto

henrybear327 (changes requested):
LGTM in general :) Great job @bruelea

pkg/netbox/api/prefix_claim.go
bruelea: It's to distinguish between the case A "validation fails" and case B "prefix length is invalid".
In case B the "isRequestingInvalidPrefixLength" the function is working as expected. 
I will change the function name and will return only an error.

pkg/netbox/api/ip_address_claim_test.go
bruelea: For IPv4 there are some test cases that where the mock is set up to return two ip addresses and some where it only returns one. As the selection of the ip uses same function for IPv4 and IPv6, I didn't add additional tests where two IPs are returned for IPv6.

henrybear327: Got it, thanks!

pkg/netbox/api/prefix_claim.go
henrybear327: Do we want to do the similar validation for IPv4

bruelea:
> LGTM in general :) Great job @bruelea

Thanks, for the feedback! @henrybear327. I edited the code accordingly and changed the error handling in the claim controllers to requeue the object if the assignment fails and not returning an error.

pkg/netbox/api/ip_address_claim_test.go
henrybear327:
```suggestion
					Family:  int64(IPv4Familiy),
```

internal/controller/prefixclaim_controller.go
```diff
@@ -105,19 +106,23 @@ func (r *PrefixClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 		h := generatePrefixRestorationHash(prefixClaim)
 		prefixModel, err := r.NetboxClient.RestoreExistingPrefixByHash(h)
 		if err != nil {
-			return ctrl.Result{}, err
+			if setConditionErr := r.SetConditionAndCreateEvent(ctx, prefixClaim, netboxv1.ConditionPrefixAssignedFalse, corev1.EventTypeWarning, err.Error()); setConditionErr != nil {
+				return ctrl.Result{}, fmt.Errorf("error updating status: %w, when look up of prefix by hash failed: %w", setConditionErr, err)
+			}
+			return ctrl.Result{Requeue: true}, nil
 		}
 		// TODO: set condition for each error
```
henrybear327, line 127:
```suggestion
```

henrybear327 (changes requested):
2 small tweaks! Thanks for addressing the comments @bruelea!

pkg/netbox/api/prefix_claim.go
bruelea: If we can select the pattern with a condition we could do the validation in the type definition. As it is the same type and field for IPv4 and IPv6 I chose the validation that includes both of them.

pkg/netbox/api/prefix_claim.go
henrybear327: Since both `err != nil` and `ret == true` will throw an error to the caller, I think it makes sense that we just throw an error from `isRequestingInvalidPrefixLength` with all the associated error messages. 

Thanks for making the changes!

pkg/netbox/api/prefix_claim.go
henrybear327: Makes sense! I checked the references it doesn't seem to allow conditional in the validation rules. 

Let's leave it as is :)

henrybear327 (approved):
LGTM! :) Thanks @bruelea 

(and nice catch on the spelling error :D)

henrybear327 (approved):

PR #78, opened 2024-09-30
Fix Finalizers

### LINKED ISSUE #77: Finalizer set too late
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via netbox operator maintainers.
- [X] Existing open issues have been checked and this is not a duplicate.

### What happened?

The finalizer is set in [line 199 of ipAddress Controller](https://github.com/netbox-community/netbox-operator/blob/main/internal/controller/ipaddress_controller.go#L199). This is after making the ipAddress Claim in Netbox. The finalizer should be set before. If now the operator crashes in between, the ip-claim will not be cleaned up. Probably should also be added before acquiring the lock of the leaselocker. The same issue [in the Prefix Controller](https://github.com/netbox-community/netbox-operator/blob/main/internal/controller/prefix_controller.go#L194)

### What did you expect to happen?

Finalizer should be added earlier, before any api calls are done which would need to be reverted in case of deletion.

### How can we reproduce it (as minimally and precisely as possible)?

Add a return call between the API call to Netbox and the addFinalizer function

### Netbox operator version

All

### Netbox operator configuration (command line flags or environment variables)

_No response_

### Relevant log output

_No response_

### Anything else we need to know?

_No response_

MaIT-HgA:
Changes:

* adding Finalizers are moved up
* IP/Prefix deletion functions are updated with not found errors
* error handling in Finalizers is updated with not found errors
* integration test is updated

Close #77

### DIFF
--- a/internal/controller/expected_netboxmock_calls_test.go
+++ b/internal/controller/expected_netboxmock_calls_test.go
@@ -139,6 +139,21 @@ func mockIpAddressesDelete(ipamMock *mock_interfaces.MockIpamInterface, catchUne
 		}).MinTimes(1)
 }
 
+func mockIpAddressesDeleteFail(ipamMock *mock_interfaces.MockIpamInterface, catchUnexpectedParams chan error) {
+	ipamMock.EXPECT().IpamIPAddressesDelete(gomock.Any(), nil).
+		DoAndReturn(func(params interface{}, authInfo interface{}, opts ...interface{}) (*ipam.IpamIPAddressesDeleteNoContent, error) {
+			got := params.(*ipam.IpamIPAddressesDeleteParams)
+			diff := deep.Equal(got, ExpectedDeleteFailParams)
+			if len(diff) > 0 {
+				err := fmt.Errorf("netboxmock: unexpected call to ipam.IpamIPAddressesDelete, diff to expected params diff: %+v", diff)
+				catchUnexpectedParams <- err
+				return &ipam.IpamIPAddressesDeleteNoContent{}, err
+			}
+			fmt.Printf("NETBOXMOCK\t ipam.IpamIPAddressesDelete was called with mock input\n")
+			return nil, ipam.NewIpamIPAddressesDeleteDefault(404)
+		}).MinTimes(1)
+}
+
 func mockIpamIPAddressesUpdate(ipamMock *mock_interfaces.MockIpamInterface, catchUnexpectedParams chan error) {
 	ipamMock.EXPECT().IpamIPAddressesUpdate(gomock.Any(), nil).
 		DoAndReturn(func(params interface{}, authInfo interface{}, opts ...interface{}) (*ipam.IpamIPAddressesUpdateOK, error) {
--- a/internal/controller/ipaddress_controller.go
+++ b/internal/controller/ipaddress_controller.go
@@ -30,6 +30,7 @@ import (
 	"github.com/netbox-community/netbox-operator/pkg/config"
 	"github.com/netbox-community/netbox-operator/pkg/netbox/api"
 	"github.com/netbox-community/netbox-operator/pkg/netbox/models"
+	"github.com/netbox-community/netbox-operator/pkg/netbox/utils"
 	"github.com/swisscom/leaselocker"
 	corev1 "k8s.io/api/core/v1"
 	apismeta "k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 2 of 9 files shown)


internal/controller/prefix_controller.go
```diff
@@ -81,6 +82,14 @@ func (r *PrefixReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		if controllerutil.ContainsFinalizer(prefix, PrefixFinalizerName) {
 			if !prefix.Spec.PreserveInNetbox {
 				if err := r.NetboxClient.DeletePrefix(prefix.Status.PrefixId); err != nil {
+					if errors.Is(err, utils.ErrNotFound) {
+						setConditionErr := r.SetConditionAndCreateEvent(ctx, prefix, netboxv1.ConditionPrefixReadyFalse, corev1.EventTypeWarning, err.Error())
```
jstudler, line 86: Propose to use a different condition (replace netboxv1.ConditionPrefixReadyFalse with a new condition like e.g. netboxv1.ConditionPrefixReadyFalseDeletionFailed)

Currently the condition looks as follows:

```
  conditions:
  - lastTransitionTime: "2024-10-01T07:14:46Z"
    message: 'Failed to reserve IP in NetBox. failed to fetch IPAddress: not found'
    reason: FailedToReserveIpInNetbox
    status: "False"
    type: Ready
```

Propose to use reason "FailedToDeleteIpInNetbox" and message "Failed to delete IP in NetBox".

Same applies to the other controller...

internal/controller/prefix_controller.go
```diff
@@ -81,6 +82,14 @@ func (r *PrefixReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		if controllerutil.ContainsFinalizer(prefix, PrefixFinalizerName) {
 			if !prefix.Spec.PreserveInNetbox {
 				if err := r.NetboxClient.DeletePrefix(prefix.Status.PrefixId); err != nil {
+					if errors.Is(err, utils.ErrNotFound) {
```
jstudler, line 85: If the Prefix is not found in NetBox, isn't this the expected target state so we should assume everything is fine and remove the finalizer? I propose to inverse the line 85 condition and remove line 93. wdyt @bruelea @MaIT-HgA ?

pkg/netbox/api/prefix.go
```diff
@@ -98,9 +100,17 @@ func (r *NetboxClient) UpdatePrefix(prefixId int64, prefix *netboxModels.Writabl
 }
 
 func (r *NetboxClient) DeletePrefix(prefixId int64) error {
+	// assuming id starts from 1 when Prefix is created so 0 means that Prefix doesn't exist
+	if prefixId == 0 {
```
jstudler, line 104: Not sure what this check is about. Is it to catch the case where the prefix was never created in NetBox?

internal/controller/prefix_controller.go
MaIT-HgA: I agree, it was a mistake from my end

internal/controller/prefix_controller.go
MaIT-HgA: I agree

bruelea (changes requested):

internal/controller/ipaddress_controller.go
```diff
@@ -81,7 +82,14 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			if !o.Spec.PreserveInNetbox {
 				err := r.NetboxClient.DeleteIpAddress(o.Status.IpAddressId)
 				if err != nil {
-					return ctrl.Result{}, err
+					if !errors.Is(err, utils.ErrNotFound) {
```
bruelea, line 85: if we return nil if we have a 404 error in DeleteIpAdress() then we don't need to check here if it's a not found error.

bruelea (changes requested):

internal/controller/expected_netboxmock_calls_test.go
```diff
@@ -139,6 +139,21 @@ func mockIpAddressesDelete(ipamMock *mock_interfaces.MockIpamInterface, catchUne
...
+		DoAndReturn(func(params interface{}, authInfo interface{}, opts ...interface{}) (*ipam.IpamIPAddressesDeleteNoContent, error) {
+			got := params.(*ipam.IpamIPAddressesDeleteParams)
+			diff := deep.Equal(got, ExpectedDeleteFailParams)
+			if len(diff) > 0 {
+				err := fmt.Errorf("netboxmock: unexpected call to ipam.IpamIPAddressesDelete, diff to expected params diff: %+v", diff)
+				catchUnexpectedParams <- err
+				return &ipam.IpamIPAddressesDeleteNoContent{}, err
+			}
+			fmt.Printf("NETBOXMOCK\t ipam.IpamIPAddressesDelete was called with mock input\n")
+			return &ipam.IpamIPAddressesDeleteNoContent{}, nil
```
bruelea, line 153:
```suggestion
			return nil, ipam.NewIpamIPAddressesDeleteDefault(404)
```

bruelea (changes requested):

bruelea (approved):

PR #81, opened 2024-10-01
Add age column and fix column order

jstudler:
A few UX changes:

- Adds an age column (with Kubebuilder this disappears from the CRD as soon as you add the first printcolumn)
- Orders the condition columns to feature the "Ready" condition as the last one

Before:

```bash
kubectl get ipc
NAME                    IPADDRESS    READY   IPASSIGNED
ipaddressclaim-sample   2.0.0.1/32   True    True
```

After:

```bash
kubectl get ipc
NAME                    IPADDRESS    IPASSIGNED   READY   AGE
ipaddressclaim-sample   2.0.0.1/32   True         True    4s
```


alexandernorth (approved):
lgtm!

Issue #79, opened 2024-09-30
Implement dynamic selection of parent prefix from a set of custom fields


henrybear327:
### What would you like to be added?

Currently, the parent prefix can only be defined from the `ParentPrefix` field in the CRD. 

We would like to extend it such that we can also use the `parentPrefixSelector` to obtain the parent prefix and perform prefix claims. 

The added selector may look something like this 
```
parentPrefixSelector:
    customFieldX: asdf
    customFieldY: wxyz
    customFieldZ: true
```

### Why is this needed?

Adding more flexibility for specifying the parent prefix, as we can pass it in by name (e.g. custom fields) instead of hard-coded values.

PR #90, opened 2024-10-08
Implement dynamic selection of parent prefix from a set of custom fields

henrybear327:
### Design

We introduce a new `parentPrefixSelector` field in the CR. It's a list of custom field key-value mapping, where both the key and value are of the type `string`. For example
```
parentPrefixSelector:
    environment: "Production"
    ipVersion: "4"
```

We now have 3 cases:
- If only `parentPrefix` is set, we continue on as-is
- If only `parentPrefixSelector` is set, we would take all Prefixes that exactly match all custom field's key-value specified in `parentPrefixSelector`.  We would then pick the first prefix that is able to satisfy the `prefixLength` requirement as the parent prefix. 
- If we have both `parentPrefix` and `parentPrefixSelector` present in the CR, the CRD validation will reject this yaml file

The parentPrefix will be set in the `status` field `ParentPrefix`, and this will be used as the source of truth for the rest of the reconcile function. 

### Known issue
- https://github.com/netbox-community/netbox-operator/issues/99

### TODO

- [x] Discuss if using `status` field to store the computed `ParentPrefix` value is desired
- [x] The restoration hash mechanism for `parentPrefixSelector` would require some documentation, as there might be the case there the picked parent prefix is different for the same `parentPrefixSelector` (e.g. due to prefix space exhaustion), and thus, the restoration might not work as user anticipated. 

### Notes

- Closes https://github.com/netbox-community/netbox-operator/issues/79
- Tests will be in a follow-up PR https://github.com/netbox-community/netbox-operator/issues/91


internal/controller/prefix_controller.go
```diff
@@ -132,9 +132,17 @@ func (r *PrefixReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			return ctrl.Result{}, err
 		}
 
+		if prefixClaim.Status.ParentPrefix == "" {
+			// the parent prefix is not computed
+			logger.Info("the parent prefix is not computed")
```
bruelea, line 137: you could set the ready condition of the prefix to false here.

internal/controller/prefixclaim_controller.go
```diff
@@ -75,7 +75,49 @@ func (r *PrefixClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				return ctrl.Result{Requeue: true}, nil
+			}
+
+			// set prefixClaim.Spec.ParentPrefix
+			// TODO(henrybear327): use best-fit algorithm to pick a parent prefix
+			parentPrefixCandidate := parentPrefixCandidates[0]
+			prefixClaim.Status.ParentPrefix = parentPrefixCandidate.Prefix
+		} else {
+			logger.Info("either ParentPrefixSelector or ParentPrefix needs to be set")
+			if err := r.SetConditionAndCreateEvent(ctx, prefixClaim, netboxv1.ConditionParentPrefixComputedFalse, corev1.EventTypeWarning, ""); err != nil {
```
bruelea, line 107: you could add the log message also to the event created with SetConditionAndCreateEvent, so that the case is visible when looking at the prefixclaim CR

internal/controller/prefix_controller.go
henrybear327: It's missed, thanks for flagging this!

internal/controller/prefixclaim_controller.go
henrybear327: I will refactor the logging part for better debugging experience :)

henrybear327:
> Discuss if using the status field to store the computed ParentPrefix value is desired

We agree that the `spec` field is read-only, so we wouldn't put any computed ParentPrefix there. 

The `status` field is what we have write access to and where we reflect the relevant internal states to the external user. Thus, we have decided to use it to store the computed ParentPrefix and use it as the source of truth.

internal/controller/prefix_controller.go
henrybear327: Fixed

henrybear327:
Improved CRD validation: According to [2], the `oneOf` validation will only come in the next version. 

cc: @alexandernorth @bruelea 

Reference:
[1] https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/
[2] https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-ratcheting

henrybear327:
For some reason, when we are querying for custom fields, for example using this URL `http://localhost:8080/ipam/prefixes/?q=&cf_poolName=Pool+2&cf_environment=production`, the value of the custom fields will be capitalized automatically. This has an impact on us as we can only enter the value in custom fields in capitalized form.

Reproduction step:
- type in `http://localhost:8080/ipam/prefixes/?q=&cf_poolName=Pool+2&cf_environment=production` in your browser
- you can see that the search that went through NetBox is actually `http://localhost:8080/ipam/prefixes/?q=&cf_poolName=Pool+2&cf_environment=Production`

The demo SQL files have adhered to this "observation", but I am not sure if this is intended or not.

cc: @alexandernorth @bruelea

internal/controller/prefixclaim_controller.go
henrybear327: The log message is merged to the event log for the case where logging happens right before the code for emitting events. See https://github.com/netbox-community/netbox-operator/pull/90/commits/44aaa8245186dea88323929950fa6008a1bcbebc

henrybear327:
To maintain the backward compatibility of the restoration hash and extend it to support parentPrefixSelection, we have made the following changes to the hash:
- append the parentPrefixSelection key-value pairs to the end of the hash during hash calculation. The exact string that is appended is done by sorting the keys first before appending, so in the case that only the key order of the parentPrefixSelection changes, the hash value won't be affected

With the aforementioned change, the hash can by design distinguish the prefix claimed by ParentPrefix or parentPrefixSelection, thus, making the restoration process relying on the hash achievable.

internal/controller/prefix_controller.go
```diff
@@ -132,9 +132,19 @@ func (r *PrefixReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			return ctrl.Result{}, err
 		}
 
+		if prefixClaim.Status.ParentPrefix == "" {
```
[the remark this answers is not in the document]
henrybear327: I am not sure if this is required, but I would like to log this in case we change more code in the future and this condition gets tripped somehow. What do you think?

internal/controller/prefixclaim_controller.go
```diff
@@ -75,7 +75,49 @@ func (r *PrefixClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				// we requeue as this might be a temporary exhausation
+				return ctrl.Result{Requeue: true}, nil
+			}
+
+			// set prefixClaim.Spec.ParentPrefix
+			// TODO(henrybear327): use best-fit algorithm to pick a parent prefix
+			parentPrefixCandidate := parentPrefixCandidates[0]
+			prefixClaim.Status.ParentPrefix = parentPrefixCandidate.Prefix
+		} else {
+			// this case should not be triggered anymore, as we have validation rules put in place on the CR
```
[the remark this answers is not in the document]
henrybear327: Same as the above, I would like to log this in case we change more code in the future and this condition gets tripped somehow. 

What do you think? Or do you think if it's a given, we don't check for it :)

pkg/netbox/api/ip_address_claim.go
```diff
@@ -40,7 +40,12 @@ const (
 )
 
 func (r *NetboxClient) RestoreExistingIpByHash(customFieldName string, hash string) (*models.IPAddress, error) {
```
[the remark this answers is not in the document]
henrybear327: Do you mean we share the hash computation formula between IP and Prefix?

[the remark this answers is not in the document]
bruelea: more the structure of the function, RestoreExistingPrefixByHash has 1 input parameter but ResoreExistingIpByHash has 2.

internal/controller/prefixclaim_controller.go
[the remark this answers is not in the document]
bruelea: we could add a test which checks that the prefix CR is not created if the ParentPrefix is not found. This way we could make sure the behavior doesn't occur until a case comes up where we need it.

api/v1/prefixclaim_types.go
```diff
@@ -56,14 +62,17 @@ type PrefixClaimSpec struct {
...
+
+	ParentPrefix string             `json:"parentPrefix,omitempty"` // Due to the fact that we can use ParentPrefixSelector to assign parent prefixes, we use this field to store exactly which parent prefix we are using for prefix allocation
+	Prefix       string             `json:"prefix,omitempty"`
+	PrefixName   string             `json:"prefixName,omitempty"`
+	Conditions   []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
 }
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="ParentPrefix",type=string,JSONPath=`.status.parentPrefix`
```
bruelea, line 75: Do we want the 'Prefix' or 'ParenPrefix' as first column? As a user the value that interests more could be the 'Prefix' that was assigned

internal/controller/prefix_controller.go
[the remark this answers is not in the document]
henrybear327: https://github.com/netbox-community/netbox-operator/pull/90#discussion_r1796932930

pkg/netbox/api/ip_address_claim.go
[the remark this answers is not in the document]
henrybear327: Nice catch. Aligned by changing the function signature to `RestoreExistingIpByHash(hash string)`

api/v1/prefixclaim_types.go
henrybear327: Nice idea. 

The new column sequence is `Prefix`, `PrefixAssigned`, `ParentPrefix`, `Ready`, `Age`

henrybear327:
> Comment from @bruelea: add the newly created sample to the `kustomization: config/samples/kustomization.yaml`

Done

jitendrs:
Please add me as reviewer.

henrybear327:
> For some reason, when we are querying for custom fields, for example using this URL `http://localhost:8080/ipam/prefixes/?q=&cf_poolName=Pool+2&cf_environment=production`, the value of the custom field `cf_environment` will be capitalized automatically. This has an impact on us as we can only enter the value in custom fields in capitalized form.
> 
> Reproduction step:
> 
> * type in `http://localhost:8080/ipam/prefixes/?q=&cf_poolName=Pool+2&cf_environment=production` in your browser
> * you can see that the search that went through NetBox is actually `http://localhost:8080/ipam/prefixes/?q=&cf_poolName=Pool+2&cf_environment=Production`
> 
> The demo SQL files have adhered to this "observation", but I am not able to wrap my head around this, since for the restoration hash, we have no issue with it
> 
> cc: @alexandernorth @bruelea

So as @alexandernorth explained, on browsers, they might modify the URLs. But for the API endpoints what we send is directly being processed. 

So in the end, this is not an issue for us when querying using API endpoints! But when we are doing debugging using URLs, we might run into issues.

api/v1/prefixclaim_types.go
```diff
@@ -56,16 +62,19 @@ type PrefixClaimSpec struct {
 type PrefixClaimStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
-	// Prefix status: container, active, reserved , deprecated
-	Prefix     string             `json:"prefix,omitempty"`
-	PrefixName string             `json:"prefixName,omitempty"`
-	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+	// Prefix status: container, active, reserved, deprecated
+
+	ParentPrefix string             `json:"parentPrefix,omitempty"` // Due to the fact that we can use ParentPrefixSelector to assign parent prefixes, we use this field to store exactly which parent prefix we are using for prefix allocation
```
jitendrs, line 67: as I understand from comment, we still need to keep referred parent prefix details, does it necessary to make it `omitempty`.

jitendrs (approved):
Great work! The changes look solid, and everything is well-structured. I believe it aligns with our expectations overall. However, the section concerning site  could be improved. I understand there is already open issue #100 for this expectation.

pkg/netbox/api/prefix_claim.go
```diff
@@ -80,6 +91,52 @@ func validatePrefixLengthOrError(prefixClaim *models.PrefixClaim, prefixFamily i
...
+					Prefix: *prefix.Prefix,
+				})
+			}
+		}
+	}
+
+	return prefixes, nil
+}
+
+func (r *NetboxClient) getAvailablePrefixByPrefix(tenant, prefix, prefixLength string) (*models.Prefix, error) {
```
CapiSSS, line 130: Maybe it's all about context, but it is not clear to me what this method does based on its name

internal/controller/prefixclaim_controller.go
```diff
@@ -75,7 +75,49 @@ func (r *PrefixClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	}
 
-	/* 1. check if the matching Prefix object exists */
+	/* 1. compute and assign the parent prefix if required */
+	// The current design will use prefixClaim.Status.ParentPrefix for storing the computed parent prefix, and as the source of truth for future parent prefix references
+	if prefixClaim.Status.ParentPrefix == "" /* parent prefix not yet computed/assigned */ {
+		if prefixClaim.Spec.ParentPrefix != "" {
+			// ParentPrefix takes precedence over ParentPrefixSelector
+			prefixClaim.Status.ParentPrefix = prefixClaim.Spec.ParentPrefix
+		} else if len(prefixClaim.Spec.ParentPrefixSelector) > 0 {
```
CapiSSS, line 84: Correct me if I'm wrong here.

This code snippet will try to get the **first** available prefix in Netbox that matches the following conditions:

- All fields set under `spec.parentPrefixSelector`
- Tenant
- (Minimum) Prefix length

What about `site`? Could it happen that the first matching available prefix is on a site different from the one specified in `spec.site`?

E.g. the first matching available prefix is from site `DataCenter2`, but in your PrefixClaim spec you specified `site: DataCenter1`. Since this value is not used as a filter for the computation of the ParentPrefix, could this be a potential issue?

CapiSSS: I guess this issue (if so) could be solved by adding `site` as a parameter inside `spec.parentPrefixSelector`. Then the only 'issue' is that it would be redundant to have the site specified in two places.

jstudler:
For parentPrefixSelector we will maybe want to add a field to the spec that specifies whether we are looking for IPv4 or IPv6. With parentPrefix it was clear since a prefix can only be IPv4 or IPv6 and so can it's childs. With parentPrefixSelector, NetBox can return both IPv6 and IPv4 prefixes. So we maybe need to add a field like addressFamily or ipVersion that should only be set if using parentPrefixSelector. Maybe we could also leverage the NetBox "family" field and just add it to parentPrefixSelector.

henrybear327:
> Question regarding the different [types of NetBox Custom Fields](https://netbox.gmedia.id/static/docs/customization/custom-fields/#creating-custom-fields): Did we verify that we support all different types (e.g. Text, Integer, Boolean) in both the .spec.parentPrefixSelector and the .spec.customFields? Maybe go-netbox already handles this for us.

The current custom field in the prefix claim CRD is a map with `<string-string>` data type. It can work normally with customs field types declared in text, integer, and boolean, as I have tested using `netbox_v1_prefixclaim_customfields_bool_int.yaml`. We can test more data types if required. 

In my observation, we just need to follow the text representation of the format for a specific data type, e.g. write `true`/`false` in quotation marks for booleans. 

---

In the `local-demo-data.sql`, several custom fields with different data types are added.

In `netbox_v1_prefixclaim_customfields`, we have the following `parentPrefixSelector`:

```
parentPrefixSelector: # notice that the keys and values are case-sensitive
    environment: "Production"
    poolName: "Pool 1"
```

In `netbox_v1_prefixclaim_customfields_bool_int.yaml`, we have the following `parentPrefixSelector`:
```
parentPrefixSelector: # notice that the keys and values are case-sensitive
    cfDataTypeBool: "true"
    cfDataTypeInteger: "1"
```

Both should be selecting the same set of parent prefixes, and it has been verified that the returned results are the same

internal/controller/prefixclaim_helpers.go
```diff
@@ -62,21 +63,44 @@ func generatePrefixSpec(claim *netboxv1.PrefixClaim, prefix string, logger logr.
...
+		Namespace:            claim.Namespace,
+		Name:                 claim.Name,
+		ParentPrefix:         claim.Status.ParentPrefix,
+		PrefixLength:         claim.Spec.PrefixLength,
+		Tenant:               claim.Spec.Tenant,
+		ParentPrefixSelector: parentPrefixSelectorStr,
 	}
-	return fmt.Sprintf("%x", sha1.Sum([]byte(rd.Namespace+rd.Name+rd.ParentPrefix+rd.PrefixLength+rd.Tenant)))
+
+	return fmt.Sprintf("%x", sha1.Sum([]byte(rd.Namespace+rd.Name+rd.ParentPrefix+rd.PrefixLength+rd.Tenant+rd.ParentPrefixSelector)))
```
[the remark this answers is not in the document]
henrybear327: Good suggestion! :) 

Applied!

pkg/netbox/api/prefix_claim.go
henrybear327: @CapiSSS  I renamed it to `checkPrefixAvailbility`. Does this help? :)

internal/controller/prefixclaim_controller.go
henrybear327: Now the computation is added to `isParentPrefixCandidate`, so all the prefix candidates have taken in `site` as a condition during filtering. 

Do you think this would be enough :)

henrybear327: I will add a new ParentPrefixSelector documenting the latest changes up there for you to review. Stayed tuned for a bit!

henrybear327:
> Great work! The changes look solid, and everything is well-structured. I believe it aligns with our expectations overall. However, the section concerning site could be improved. I understand there is already open issue #100 for this expectation.

Site is now made to be immutable per https://github.com/netbox-community/netbox-operator/issues/100. 

cc: @jitendrs

internal/controller/prefixclaim_controller.go
henrybear327: Please see `ParentPrefixSelectorGuide.md`!

api/v1/prefixclaim_types.go
```diff
@@ -24,24 +24,31 @@ import (
...
 	// Important: Run "make" to regenerate code after modifying this file
 
-	//+kubebuilder:validation:Required
 	//+kubebuilder:validation:Format=cidr
 	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'parentPrefix' is immutable"
-	ParentPrefix string `json:"parentPrefix"`
+	ParentPrefix string `json:"parentPrefix,omitempty"`
+
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'parentPrefixSelector' is immutable"
+	ParentPrefixSelector map[string]string `json:"parentPrefixSelector,omitempty"`
```
jstudler, line 40: Please add as much documentation as possible in here. This way it will appear in the API Schema which will be viewable using e.g. `kubectl explain`. Things such as the following should be mentioned

- You can use both custom fields and built in fields
- For tenant and site use the name instead of the slug

The markdown files should be complementary (e.g. with the example or further reading)

config/samples/netbox_v1_prefixclaim_customfields_bool_int.yaml
jstudler, line 1: Same here. Please don't use customfields in the filename.

config/samples/netbox_v1_prefixclaim_customfields.yaml
```diff
@@ -0,0 +1,19 @@
...
+spec:
+  tenant: "MY_TENANT" # Use the `name` value instead of the `slug` value
+  site: "DM-Akron" # Use the `name` value instead of the `slug` value
+  description: "some description"
+  comments: "your comments"
+  preserveInNetbox: true
+  prefixLength: "/31"
+  parentPrefixSelector: # The keys and values are case-sensitive
+    environment: "Production"
+    poolName: "Pool 1"
```
jstudler, line 17: A bit confused here. The custom field name is `poolName`. Shouldn't the filter in here then be `cf_poolName`? Or how do you distinguish custom fields from built in fields? I think it would make sense to be consistent with the netbox filtering https://demo.netbox.dev/static/docs/rest-api/filtering/#filtering-by-custom-field

pkg/netbox/api/helper.go
```diff
@@ -24,24 +24,46 @@ import (
 	"github.com/go-openapi/strfmt"
 )
 
-type CustomFieldStringFilter struct {
-	CustomFieldName  string
-	CustomFieldValue string
+type CustomFieldEntry struct {
```
jstudler, line 27: Is `CustomField` the right wording here or should it rather be `Filter`?

pkg/netbox/api/helper.go
```diff
@@ -24,24 +24,46 @@ import (
...
+	value string
 }
 
-func newCustomFieldStringFilterOperation(name string, value string) func(co *runtime.ClientOperation) {
+type QueryFieldStringFilter struct {
+	fields       map[string]string
+	customFields []CustomFieldEntry
+}
+
+func newQueryFieldStringFilterOperation(fields map[string]string, customFields []CustomFieldEntry) func(co *runtime.ClientOperation) {
```
jstudler, line 37: Is `customFields` the right wording here or should it rather be `filters`?

jstudler (changes requested):
Hey @henrybear327 
Thanks for the work. I did some black box testing without digging too much into the code so far. I've left some comments regarding wording and documentation. Please remove "custom field" where it's not explicitly custom fields to avoid confusion for users. 

Manual testing on a local kind cluster didn't work for me:

```
make create-kind && make deploy-kind
kubectl apply -f config/samples/netbox_v1_prefixclaim_customfields.yaml
kubectl describe pxc prefixclaim-customfields-sample
Name:         prefixclaim-customfields-sample
Namespace:    default
Labels:       app.kubernetes.io/managed-by=kustomize
              app.kubernetes.io/name=netbox-operator
Annotations:  <none>
API Version:  netbox.dev/v1
Kind:         PrefixClaim
Metadata:
  Creation Timestamp:  2024-11-12T08:52:02Z
  Generation:          1
  Resource Version:    1495
  UID:                 ab6c76d1-08b4-4a2d-894c-da04f3961b4e
Spec:
  Comments:     your comments
  Description:  some description
  Parent Prefix Selector:
    Environment:       Production
    Pool Name:         Pool 1
  Prefix Length:       /31
  Preserve In Netbox:  true
  Site:                DM-Akron
  Tenant:              MY_TENANT
Events:                <none>
```

And the operator logs are rather short:

```
2024-11-12T09:02:57Z    INFO    prefixClaim reconcile loop started      {"controller": "prefixclaim", "controllerGroup": "netbox.dev", "controllerKind": "PrefixClaim", "PrefixClaim": {"name":"prefixclaim-customfields-sample","namespace":"default"}, "namespace": "default", "name": "prefixclaim-customfields-sample", "reconcileID": "2b0e902f-aae0-42a4-860e-f96f0aba8ea3"}
2024-11-12T09:02:57Z    ERROR   Reconciler error        {"controller": "prefixclaim", "controllerGroup": "netbox.dev", "controllerKind": "PrefixClaim", "PrefixClaim": {"name":"prefixclaim-customfields-sample","namespace":"default"}, "namespace": "default", "name": "prefixclaim-customfields-sample", "reconcileID": "2b0e902f-aae0-42a4-860e-f96f0aba8ea3", "error": "resource name may not be empty"}
```

Will try again later with running it locally against demo.netbox.dev.

ParentPrefixSelectorGuide.md
```diff
@@ -0,0 +1,51 @@
+# A guide of `ParentPrefixSelector` in `PrefixClaim`
+
+There are 2 ways to make a Prefix claim:
+- provide a `parentPrefix`
+- provide a `parentPrefixSelector`
+
+In this documentation, we will focus on the `parentPrefixSelector` only.
```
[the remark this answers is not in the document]
henrybear327: > Please add a note on how it behaves in case both are set. Which one takes precedence?

In the current design, CRD validation will reject on applying.

config/samples/netbox_v1_prefixclaim_customfields.yaml
henrybear327: I manually filter out `tenant` and `site`, so we don't force the user to build the netbox API query string in the CRs. 

Should we have the user enter the prefix `cf_` when adding the entries to save us some implementation hassle, or, we handle this for the user, which looks nicer on the UX standpoint.  What do you think?

jstudler:
If you use a PrefixClaim using parentPrefixSelector with an invalid Tenant (a Tenant that doesn't exist in NetBox), the Event will show: 

```
The parent prefix was not able to be computed. no parent prefix can be obtained with the query conditions set in ParentPrefixSelector, err = <nil>, number of candidates = 0
```

Spec example:

```yaml
apiVersion: netbox.dev/v1
kind: PrefixClaim
metadata:
  name: prefixclaim-customfields-sample
spec:
  tenant: "MY_TENANT that doesn't exist"
  site: "DM-Akron"
  preserveInNetbox: true
  prefixLength: "/31"
  parentPrefixSelector: # The keys and values are case-sensitive
    environment: "PostProduction"
    poolName: "Pool 1"
```

This is quite misleading. Do you think there is an easy way to enhance this?

jstudler:
Under some circumstances the parent prefix will be computed but it's already exhausted. Even if you add more prefixes, the system will not recover and compute/chose a new parent prefix. Example events:

```
Events:
  Type     Reason                    Age    From                     Message
  ----     ------                    ----   ----                     -------
  Normal   ParentPrefixComputed      4m41s  prefix-claim-controller  The parent prefix was computed successfully. parentPrefix is computed: 1.254.0.0/27
  Warning  PrefixCRNotCreated        3m39s  prefix-claim-controller  Failed to fetch new Prefix from NetBox. parent prefix exhausted
  Warning  FailedToLockParentPrefix  2m     prefix-claim-controller  failed to lock parent prefix 1.254.0.0/27
```


Could we fix this by removing `.status.parentPrefix` if we come to the condition that has the message "parent prefix exhausted"?

api/v1/prefixclaim_types.go
```diff
@@ -58,16 +65,19 @@ type PrefixClaimSpec struct {
...
+	Prefix       string             `json:"prefix,omitempty"`
+	PrefixName   string             `json:"prefixName,omitempty"`
+	Conditions   []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
 }
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Prefix",type=string,JSONPath=`.status.prefix`
 // +kubebuilder:printcolumn:name="PrefixAssigned",type=string,JSONPath=`.status.conditions[?(@.type=="PrefixAssigned")].status`
+// +kubebuilder:printcolumn:name="ParentPrefix",type=string,JSONPath=`.status.parentPrefix`
```
jstudler, line 80: Not sure if we want to have this print column. If we have parentPrefixSelector, this should be `.status.parentPrefix`, otherwise it should be `.spec.parentPrefix`. I'm in favor of removing it to avoid user confusion.

api/v1/prefixclaim_types.go
```diff
@@ -58,16 +65,19 @@ type PrefixClaimSpec struct {
 type PrefixClaimStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
-	// Prefix status: container, active, reserved , deprecated
-	Prefix     string             `json:"prefix,omitempty"`
-	PrefixName string             `json:"prefixName,omitempty"`
-	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+	// Prefix status: container, active, reserved, deprecated
+
+	ParentPrefix string             `json:"parentPrefix,omitempty"` // Due to the fact that we can use ParentPrefixSelector to assign parent prefixes, we use this field to store exactly which parent prefix we are using for prefix allocation
```
jstudler, line 70: Propose to change this to `SelectedParendPrefix` to make it relate to `ParentPrefixSelector`

api/v1/prefixclaim_types.go
```diff
@@ -120,3 +130,17 @@ var ConditionPrefixAssignedFalse = metav1.Condition{
 	Reason:  "PrefixCRNotCreated",
 	Message: "Failed to fetch new Prefix from NetBox",
 }
+
+var ConditionParentPrefixComputedTrue = metav1.Condition{
```
jstudler, line 134: To match with the wording of `ParentPrefixSelector`, could we maybe use `Selected` instead of `Computed`?

api/v1/prefixclaim_types.go
henrybear327: Renamed

api/v1/prefixclaim_types.go
henrybear327: All the `computed` occurrences are replaced with `selected`

pkg/netbox/api/helper.go
henrybear327: I think `CustomField` is a better name. 

This structure originated from the need to query the restoration hash stored within the custom field. 

We are leveraging this design by enabling querying multiple custom fields to support the implementation of dynamic parent prefixes.

config/samples/netbox_v1_prefixclaim_customfields.yaml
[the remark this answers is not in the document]
henrybear327: I keep the tradition of using all lowercase characters in the filename!

config/samples/netbox_v1_prefixclaim_customfields.yaml
```diff
@@ -0,0 +1,19 @@
...
+    # environment: "production"
+    # poolName: "pool 3"
```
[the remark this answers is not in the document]
henrybear327: Added!

The CR in `ParentPrefixSelectorGuide.md` is also updated accordingly. Thanks!

api/v1/prefixclaim_types.go
henrybear327: Done as requested, also updated the description in `ParentPrefixSelectorGuide.md`.

henrybear327: FYI, I have created https://github.com/netbox-community/netbox-operator/issues/128 to track for other fields and CRDs

henrybear327:
@jstudler please take another pass when you have time! :) Thanks

henrybear327:
> If you use a PrefixClaim using parentPrefixSelector with an invalid Tenant (a Tenant that doesn't exist in NetBox), the Event will show:
> 
> ```
> The parent prefix was not able to be computed. no parent prefix can be obtained with the query conditions set in ParentPrefixSelector, err = <nil>, number of candidates = 0
> ```
> 
> Spec example:
> 
> ```yaml
> apiVersion: netbox.dev/v1
> kind: PrefixClaim
> metadata:
>   name: prefixclaim-customfields-sample
> spec:
>   tenant: "MY_TENANT that doesn't exist"
>   site: "DM-Akron"
>   preserveInNetbox: true
>   prefixLength: "/31"
>   parentPrefixSelector: # The keys and values are case-sensitive
>     environment: "PostProduction"
>     poolName: "Pool 1"
> ```
> 
> This is quite misleading. Do you think there is an easy way to enhance this?

We actually throw an error indicating this situation, please see the code snippet below:
```go
func (r *NetboxClient) GetTenantDetails(name string) (*models.Tenant, error) {
	request := tenancy.NewTenancyTenantsListParams().WithName(&name)
	response, err := r.Tenancy.TenancyTenantsList(request, nil)
	if err != nil {
		return nil, utils.NetboxError("failed to fetch Tenant details", err)
	}
	if len(response.Payload.Results) == 0 {
		return nil, utils.NetboxNotFoundError("tenant '" + name + "'")
	}

	return &models.Tenant{
		Id:   response.Payload.Results[0].ID,
		Slug: *response.Payload.Results[0].Slug,
		Name: *response.Payload.Results[0].Name,
	}, nil
}
```

I think this is related to the discussion this morning - we need to surface the underlying errors properly to the user. As in this case, we did throw an error clearly specifying the problem, but it's not visible to the user.

henrybear327:
> Under some circumstances the parent prefix will be computed but it's already exhausted. Even if you add more prefixes, the system will not recover and compute/chose a new parent prefix. Example describe:
> 
> ```
> Name:         prefixclaim-customfields-sample-09
> Namespace:    default
> Labels:       <none>
> Annotations:  <none>
> API Version:  netbox.dev/v1
> Kind:         PrefixClaim
> Metadata:
>   Creation Timestamp:  2024-11-12T14:26:55Z
>   Generation:          1
>   Resource Version:    4212
>   UID:                 a3b68d89-6c66-4d63-8b39-c95dd9c326b5
> Spec:
>   Parent Prefix Selector:
>     Environment:       PostProduction
>     Pool Name:         Pool 254
>   Prefix Length:       /30
>   Preserve In Netbox:  false
>   Tenant:              Dunder-Mifflin, Inc.
> Status:
>   Conditions:
>     Last Transition Time:  2024-11-12T14:27:11Z
>     Message:               The parent prefix was computed successfully. parentPrefix is computed: 1.254.0.0/27
>     Reason:                ParentPrefixComputed
>     Status:                True
>     Type:                  ParentPrefixComputed
>     Last Transition Time:  2024-11-12T14:28:13Z
>     Message:               Failed to fetch new Prefix from NetBox. parent prefix exhausted
>     Reason:                PrefixCRNotCreated
>     Status:                False
>     Type:                  PrefixAssigned
>   Parent Prefix:           1.254.0.0/27
> Events:
>   Type     Reason                    Age    From                     Message
>   ----     ------                    ----   ----                     -------
>   Normal   ParentPrefixComputed      4m41s  prefix-claim-controller  The parent prefix was computed successfully. parentPrefix is computed: 1.254.0.0/27
>   Warning  PrefixCRNotCreated        3m39s  prefix-claim-controller  Failed to fetch new Prefix from NetBox. parent prefix exhausted
>   Warning  FailedToLockParentPrefix  2m     prefix-claim-controller  failed to lock parent prefix 1.254.0.0/27
> ```
> 
> Could we fix this by removing `.status.parentPrefix` if we come to the condition that has the message "parent prefix exhausted"?

Quick question, what's the desired behavior? 

IMO, prefix exhaustion might be a temporary issue, so we should requeue and keep trying.

WDYT?

ParentPrefixSelectorGuide.md
```diff
@@ -0,0 +1,56 @@
...
+
+The map contains a set of query conditions for selecting a set of prefixes that can be used as the parent prefix.
+
+The query conditions will be chained by the AND operator, and exact match of the keys and values will be performed.
+
+The fields that can be used as query conditions in the `parentPrefixSelector` are:
+- `tenant` and `site` (in lowercase characters)
+    - these 2 fields are built-in fields from NetBox, so you do *not* need to create custom fields for them
+    - please provide the *name*, not the *slug* value
+    - if the entry for tenent or site is missing, it will *not* inherit from the tenant and site from the Spec
```
CapiSSS, line 53:
```suggestion
    - if the entry for tenant or site is missing, it will *not* inherit from the tenant and site from the Spec
```

CapiSSS (approved):

api/v1/prefixclaim_types.go
jstudler: looks good, thanks!

config/samples/netbox_v1_prefixclaim_customfields.yaml
jstudler: I'm fine both ways. If we handle `cf_` for the user it's one one side nice but on the other side also confusing. Depends a bit whether we want to stick to the UI (handle it for the user) or API (the user has to specify `cf_` himself). wdyt @bruelea ?

pkg/netbox/api/helper.go
jstudler: Makes sense.

pkg/netbox/api/helper.go
jstudler: Looks good, thanks!

jstudler:
> > Under some circumstances the parent prefix will be computed but it's already exhausted. Even if you add more prefixes, the system will not recover and compute/chose a new parent prefix. Example describe:
> > ```
> > Name:         prefixclaim-customfields-sample-09
> > Namespace:    default
> > Labels:       <none>
> > Annotations:  <none>
> > API Version:  netbox.dev/v1
> > Kind:         PrefixClaim
> > Metadata:
> >   Creation Timestamp:  2024-11-12T14:26:55Z
> >   Generation:          1
> >   Resource Version:    4212
> >   UID:                 a3b68d89-6c66-4d63-8b39-c95dd9c326b5
> > Spec:
> >   Parent Prefix Selector:
> >     Environment:       PostProduction
> >     Pool Name:         Pool 254
> >   Prefix Length:       /30
> >   Preserve In Netbox:  false
> >   Tenant:              Dunder-Mifflin, Inc.
> > Status:
> >   Conditions:
> >     Last Transition Time:  2024-11-12T14:27:11Z
> >     Message:               The parent prefix was computed successfully. parentPrefix is computed: 1.254.0.0/27
> >     Reason:                ParentPrefixComputed
> >     Status:                True
> >     Type:                  ParentPrefixComputed
> >     Last Transition Time:  2024-11-12T14:28:13Z
> >     Message:               Failed to fetch new Prefix from NetBox. parent prefix exhausted
> >     Reason:                PrefixCRNotCreated
> >     Status:                False
> >     Type:                  PrefixAssigned
> >   Parent Prefix:           1.254.0.0/27
> > Events:
> >   Type     Reason                    Age    From                     Message
> >   ----     ------                    ----   ----                     -------
> >   Normal   ParentPrefixComputed      4m41s  prefix-claim-controller  The parent prefix was computed successfully. parentPrefix is computed: 1.254.0.0/27
> >   Warning  PrefixCRNotCreated        3m39s  prefix-claim-controller  Failed to fetch new Prefix from NetBox. parent prefix exhausted
> >   Warning  FailedToLockParentPrefix  2m     prefix-claim-controller  failed to lock parent prefix 1.254.0.0/27
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could we fix this by removing `.status.parentPrefix` if we come to the condition that has the message "parent prefix exhausted"?
> 
> Quick question, what's the desired behavior?
> 
> IMO, prefix exhaustion might be a temporary issue, so we should requeue and keep trying.
> 
> WDYT?

The desired behavior is that the Operator can recover from this case. Yes, we should assume that prefix exhaustion is a temporary issue only. The behaviour I've seen above was deadlock: The parent prefix was chosen but it had no space left, but there were other parent prefix candidates that weren't chosen. So I think removing `.status.parentPrefix` where the controllers writes the condition/event that contains the message "parent prefix exhausted" should fix this.

jstudler:
> > If you use a PrefixClaim using parentPrefixSelector with an invalid Tenant (a Tenant that doesn't exist in NetBox), the Event will show:
> > ```
> > The parent prefix was not able to be computed. no parent prefix can be obtained with the query conditions set in ParentPrefixSelector, err = <nil>, number of candidates = 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Spec example:
> > ```yaml
> > apiVersion: netbox.dev/v1
> > kind: PrefixClaim
> > metadata:
> >   name: prefixclaim-customfields-sample
> > spec:
> >   tenant: "MY_TENANT that doesn't exist"
> >   site: "DM-Akron"
> >   preserveInNetbox: true
> >   prefixLength: "/31"
> >   parentPrefixSelector: # The keys and values are case-sensitive
> >     environment: "PostProduction"
> >     poolName: "Pool 1"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is quite misleading. Do you think there is an easy way to enhance this?
> 
> We actually throw an error indicating this situation, please see the code snippet below:
> 
> ```go
> func (r *NetboxClient) GetTenantDetails(name string) (*models.Tenant, error) {
> 	request := tenancy.NewTenancyTenantsListParams().WithName(&name)
> 	response, err := r.Tenancy.TenancyTenantsList(request, nil)
> 	if err != nil {
> 		return nil, utils.NetboxError("failed to fetch Tenant details", err)
> 	}
> 	if len(response.Payload.Results) == 0 {
> 		return nil, utils.NetboxNotFoundError("tenant '" + name + "'")
> 	}
> 
> 	return &models.Tenant{
> 		Id:   response.Payload.Results[0].ID,
> 		Slug: *response.Payload.Results[0].Slug,
> 		Name: *response.Payload.Results[0].Name,
> 	}, nil
> }
> ```
> 
> I think this is related to the discussion this morning - we need to surface the underlying errors properly to the user. As in this case, we did throw an error clearly specifying the problem, but it's not visible to the user.

But I think the event is wrong. The event indicates "The parent prefix was not able to be computed" when in fact the problem is that the user set an invalid tenant. Could we create an event in `GetTenantDetails` for the case that the tenant cannot be found?

ParentPrefixSelectorGuide.md
henrybear327: Thanks for the catch

config/samples/netbox_v1_prefixclaim_customfields.yaml
bruelea: For spec.customFields the `cf_` is handled for the use, so we could do the same for parentPrefixSelector.

internal/controller/prefixclaim_controller.go
```diff
@@ -75,7 +79,101 @@ func (r *PrefixClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			}
+		} else if len(prefixClaim.Spec.ParentPrefixSelector) > 0 {
+			// we first check if a prefix can be restored from the netbox
+
+			// since the parent prefix is not part of the restoration hash computation
+			// we can quickly check to see if the prefix with the restoration hash is matched in NetBox
+			h := generatePrefixRestorationHash(prefixClaim)
+			canBeRestored, err := r.NetboxClient.RestoreExistingPrefixByHash(h)
+			if err != nil {
+				return ctrl.Result{Requeue: true}, nil
```
bruelea, line 102: Here the log message/condition update is missing in case of an error.

henrybear327: Added as requested! Can you see if this is what you are thinking about :)
Thanks

henrybear327:
> > > If you use a PrefixClaim using parentPrefixSelector with an invalid Tenant (a Tenant that doesn't exist in NetBox), the Event will show:
> > > ```
> > > The parent prefix was not able to be computed. no parent prefix can be obtained with the query conditions set in ParentPrefixSelector, err = <nil>, number of candidates = 0
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Spec example:
> > > ```yaml
> > > apiVersion: netbox.dev/v1
> > > kind: PrefixClaim
> > > metadata:
> > >   name: prefixclaim-customfields-sample
> > > spec:
> > >   tenant: "MY_TENANT that doesn't exist"
> > >   site: "DM-Akron"
> > >   preserveInNetbox: true
> > >   prefixLength: "/31"
> > >   parentPrefixSelector: # The keys and values are case-sensitive
> > >     environment: "PostProduction"
> > >     poolName: "Pool 1"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This is quite misleading. Do you think there is an easy way to enhance this?
> > 
> > 
> > We actually throw an error indicating this situation, please see the code snippet below:
> > ```go
> > func (r *NetboxClient) GetTenantDetails(name string) (*models.Tenant, error) {
> > 	request := tenancy.NewTenancyTenantsListParams().WithName(&name)
> > 	response, err := r.Tenancy.TenancyTenantsList(request, nil)
> > 	if err != nil {
> > 		return nil, utils.NetboxError("failed to fetch Tenant details", err)
> > 	}
> > 	if len(response.Payload.Results) == 0 {
> > 		return nil, utils.NetboxNotFoundError("tenant '" + name + "'")
> > 	}
> > 
> > 	return &models.Tenant{
> > 		Id:   response.Payload.Results[0].ID,
> > 		Slug: *response.Payload.Results[0].Slug,
> > 		Name: *response.Payload.Results[0].Name,
> > 	}, nil
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this is related to the discussion this morning - we need to surface the underlying errors properly to the user. As in this case, we did throw an error clearly specifying the problem, but it's not visible to the user.
> 
> But I think the event is wrong. The event indicates "The parent prefix was not able to be computed" when in fact the problem is that the user set an invalid tenant. Could we create an event in `GetTenantDetails` for the case that the tenant cannot be found?

Tracking under bug https://github.com/netbox-community/netbox-operator/issues/132

henrybear327:
> > > Under some circumstances the parent prefix will be computed but it's already exhausted. Even if you add more prefixes, the system will not recover and compute/chose a new parent prefix. Example describe:
> > > ```
> > > Name:         prefixclaim-customfields-sample-09
> > > Namespace:    default
> > > Labels:       <none>
> > > Annotations:  <none>
> > > API Version:  netbox.dev/v1
> > > Kind:         PrefixClaim
> > > Metadata:
> > >   Creation Timestamp:  2024-11-12T14:26:55Z
> > >   Generation:          1
> > >   Resource Version:    4212
> > >   UID:                 a3b68d89-6c66-4d63-8b39-c95dd9c326b5
> > > Spec:
> > >   Parent Prefix Selector:
> > >     Environment:       PostProduction
> > >     Pool Name:         Pool 254
> > >   Prefix Length:       /30
> > >   Preserve In Netbox:  false
> > >   Tenant:              Dunder-Mifflin, Inc.
> > > Status:
> > >   Conditions:
> > >     Last Transition Time:  2024-11-12T14:27:11Z
> > >     Message:               The parent prefix was computed successfully. parentPrefix is computed: 1.254.0.0/27
> > >     Reason:                ParentPrefixComputed
> > >     Status:                True
> > >     Type:                  ParentPrefixComputed
> > >     Last Transition Time:  2024-11-12T14:28:13Z
> > >     Message:               Failed to fetch new Prefix from NetBox. parent prefix exhausted
> > >     Reason:                PrefixCRNotCreated
> > >     Status:                False
> > >     Type:                  PrefixAssigned
> > >   Parent Prefix:           1.254.0.0/27
> > > Events:
> > >   Type     Reason                    Age    From                     Message
> > >   ----     ------                    ----   ----                     -------
> > >   Normal   ParentPrefixComputed      4m41s  prefix-claim-controller  The parent prefix was computed successfully. parentPrefix is computed: 1.254.0.0/27
> > >   Warning  PrefixCRNotCreated        3m39s  prefix-claim-controller  Failed to fetch new Prefix from NetBox. parent prefix exhausted
> > >   Warning  FailedToLockParentPrefix  2m     prefix-claim-controller  failed to lock parent prefix 1.254.0.0/27
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Could we fix this by removing `.status.parentPrefix` if we come to the condition that has the message "parent prefix exhausted"?
> > 
> > 
> > Quick question, what's the desired behavior?
> > IMO, prefix exhaustion might be a temporary issue, so we should requeue and keep trying.
> > WDYT?
> 
> The desired behavior is that the Operator can recover from this case. Yes, we should assume that prefix exhaustion is a temporary issue only. The behaviour I've seen above was deadlock: The parent prefix was chosen but it had no space left, but there were other parent prefix candidates that weren't chosen. So I think removing `.status.parentPrefix` where the controllers writes the condition/event that contains the message "parent prefix exhausted" should fix this.

Tracking under bug https://github.com/netbox-community/netbox-operator/issues/131

henrybear327:
> > > > Under some circumstances the parent prefix will be computed but it's already exhausted. Even if you add more prefixes, the system will not recover and compute/chose a new parent prefix. Example describe:
> > > > ```
> > > > Name:         prefixclaim-customfields-sample-09
> > > > Namespace:    default
> > > > Labels:       <none>
> > > > Annotations:  <none>
> > > > API Version:  netbox.dev/v1
> > > > Kind:         PrefixClaim
> > > > Metadata:
> > > >   Creation Timestamp:  2024-11-12T14:26:55Z
> > > >   Generation:          1
> > > >   Resource Version:    4212
> > > >   UID:                 a3b68d89-6c66-4d63-8b39-c95dd9c326b5
> > > > Spec:
> > > >   Parent Prefix Selector:
> > > >     Environment:       PostProduction
> > > >     Pool Name:         Pool 254
> > > >   Prefix Length:       /30
> > > >   Preserve In Netbox:  false
> > > >   Tenant:              Dunder-Mifflin, Inc.
> > > > Status:
> > > >   Conditions:
> > > >     Last Transition Time:  2024-11-12T14:27:11Z
> > > >     Message:               The parent prefix was computed successfully. parentPrefix is computed: 1.254.0.0/27
> > > >     Reason:                ParentPrefixComputed
> > > >     Status:                True
> > > >     Type:                  ParentPrefixComputed
> > > >     Last Transition Time:  2024-11-12T14:28:13Z
> > > >     Message:               Failed to fetch new Prefix from NetBox. parent prefix exhausted
> > > >     Reason:                PrefixCRNotCreated
> > > >     Status:                False
> > > >     Type:                  PrefixAssigned
> > > >   Parent Prefix:           1.254.0.0/27
> > > > Events:
> > > >   Type     Reason                    Age    From                     Message
> > > >   ----     ------                    ----   ----                     -------
> > > >   Normal   ParentPrefixComputed      4m41s  prefix-claim-controller  The parent prefix was computed successfully. parentPrefix is computed: 1.254.0.0/27
> > > >   Warning  PrefixCRNotCreated        3m39s  prefix-claim-controller  Failed to fetch new Prefix from NetBox. parent prefix exhausted
> > > >   Warning  FailedToLockParentPrefix  2m     prefix-claim-controller  failed to lock parent prefix 1.254.0.0/27
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Could we fix this by removing `.status.parentPrefix` if we come to the condition that has the message "parent prefix exhausted"?
> > > 
> > > 
> > > Quick question, what's the desired behavior?
> > > IMO, prefix exhaustion might be a temporary issue, so we should requeue and keep trying.
> > > WDYT?
> > 
> > 
> > The desired behavior is that the Operator can recover from this case. Yes, we should assume that prefix exhaustion is a temporary issue only. The behaviour I've seen above was deadlock: The parent prefix was chosen but it had no space left, but there were other parent prefix candidates that weren't chosen. So I think removing `.status.parentPrefix` where the controllers writes the condition/event that contains the message "parent prefix exhausted" should fix this.
> 
> Tracking under bug #131

Fixed. Please take a look @jstudler, thanks

config/samples/netbox_v1_prefixclaim_customfields.yaml
henrybear327: Thanks @jstudler and @bruelea for the discussion. Closing as we have reached the conclusion to keep the existing behavior.

pkg/netbox/api/prefix_claim.go
henrybear327: Thanks for checking @CapiSSS!

internal/controller/prefixclaim_controller.go
henrybear327: Thanks for checking @CapiSSS

jstudler (approved):
lgtm. Thanks a lot for the work @henrybear327 !

config/samples/kustomization.yaml
```diff
@@ -4,4 +4,5 @@ resources:
 - netbox_v1_ipaddressclaim.yaml
 - netbox_v1_prefix.yaml
 - netbox_v1_prefixclaim.yaml
+- netbox_v1_prefixclaim_customfields.yaml
```
bruelea, line 7:
```suggestion
- netbox_v1_prefixclaim_parentprefixselector_bool_int.yaml
- netbox_v1_prefixclaim_parentprefixselector.yaml
```
Just noticed that the filenames changed :)

henrybear327: Good catch! :) Thanks

bruelea (approved):

PR #107, opened 2024-10-25
Fix: add logic to update site of prefix in netbox

bruelea:
Currently, the site that is set in a prefixclaim CR is not updated in NetBox. This PR focuses on fixing the underlying logic to ensure accurate updates.

Using the go-netbox client the site is not removed from the resource in NetBox with the IpamPrefixesUpdate function. To avoid inconsistent state between the Prefix CR and the Prefix Resource in NetBox a validation rule was added to the prefix type which does not allow the deletion of the site field once it was created.

### DIFF
--- a/api/v1/prefix_types.go
+++ b/api/v1/prefix_types.go
@@ -24,6 +24,7 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // PrefixSpec defines the desired state of Prefix
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.site) || has(self.site)", message="Site is required once set"
 type PrefixSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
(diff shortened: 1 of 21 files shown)


api/v1/prefix_types.go
```diff
@@ -24,6 +24,7 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // PrefixSpec defines the desired state of Prefix
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.site) || has(self.site)", message="Site is required once set"
```
henrybear327, line 27: Out of curiosity, can you help me understand why we need to do validation here and then once more at `Site` on Line 37? :)

bruelea: Certainly! The first check, at Line 27, ensures that the field actually exists in the data. The second check, at Line 37, ensures that the field is not an empty string. The second step is only performed if the field exists, so if the first validation isn't present the second check could be skipped.

henrybear327 (approved):

PR #130, opened 2024-11-15
Support reservation of IpRanges

bruelea:
Implements https://github.com/netbox-community/netbox-operator/issues/137


internal/controller/iprange_controller.go
```diff
@@ -0,0 +1,338 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	netboxIpRangeModel, err := r.NetboxClient.ReserveOrUpdateIpRange(ipRangeModel)
+	if err != nil {
+		err := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpRangeReadyFalse,
+			corev1.EventTypeWarning, fmt.Sprintf("%s-%s", o.Spec.StartAddress, o.Spec.EndAddress), err)
+		if err != nil {
+			return ctrl.Result{}, nil
```
henrybear327, line 124: Quick question, is there a reason that we are ignoring the error here? :)

henrybear327 (changes requested):

internal/controller/iprange_controller.go
bruelea: Thanks for leaving a note! I overlooked this one, changed it to return the error if the status update fails.

henrybear327: :)

internal/controller/iprangeclaim_controller.go
```diff
@@ -0,0 +1,323 @@
...
+	err := r.Client.Get(ctx, req.NamespacedName, o)
+	if err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// if being deleted
+	if !o.ObjectMeta.DeletionTimestamp.IsZero() {
+		// end loop if deletion timestamp is not zero
+		return ctrl.Result{}, nil
+	}
```
alexandernorth, line 75: Are we missing logic for finalizers? we should only remove the claim if there is no associated range (because it has e.g. been deleted already)

internal/controller/iprangeclaim_controller.go
```diff
@@ -0,0 +1,323 @@
...
+
+	//  check if matching IpRange object already exists
+	ipRange := &netboxv1.IpRange{}
+	ipRangeName := o.ObjectMeta.Name
```
alexandernorth, line 79: like with Namespace below I would do `o.Name` for readability

internal/controller/iprangeclaim_controller.go
```diff
@@ -0,0 +1,323 @@
...
+			return ctrl.Result{}, err
+		}
+
+	} else {
+		// update fields of IPRange object
+		logger.V(4).Info("update iprange resource")
+		err := r.Client.Get(ctx, ipRangeLookupKey, ipRange)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
alexandernorth, line 127: Why is there a second `Get`? Nothing should have changed here as we are in the 'nil error' part of the branch, i.e. we got `ipRange`

internal/controller/iprangeclaim_controller.go
```diff
@@ -0,0 +1,323 @@
...
+			// only add the mutable fields here
+			ipRange.Spec.CustomFields = updatedIpRangeSpec.CustomFields
+			ipRange.Spec.Comments = updatedIpRangeSpec.Comments
+			ipRange.Spec.Description = updatedIpRangeSpec.Description
+			ipRange.Spec.PreserveInNetbox = updatedIpRangeSpec.PreserveInNetbox
+			return nil
+		})
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
alexandernorth, line 140: I would either use this CreateOrUpdate and remove the branching logic, or explicitly use the `Create` and `Update` functions.

internal/controller/iprangeclaim_controller.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	if apismeta.IsStatusConditionTrue(ipRange.Status.Conditions, "Ready") {
+		logger.V(4).Info("iprange status ready true")
+		o.Status, err = r.generateIpRangeClaimStatus(o, ipRange)
+		if err != nil {
+			logger.Error(err, "failed ot generate ip range status")
```
alexandernorth, line 147:
```suggestion
			logger.Error(err, "failed to generate ip range status")
```

internal/controller/iprangeclaim_controller.go
```diff
@@ -0,0 +1,323 @@
...
+			return nil
+		})
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	if apismeta.IsStatusConditionTrue(ipRange.Status.Conditions, "Ready") {
+		logger.V(4).Info("iprange status ready true")
+		o.Status, err = r.generateIpRangeClaimStatus(o, ipRange)
```
alexandernorth, line 145: there is no call to update after this to store the new status (unless this is achieved by setting the condition, but then this is a 'happy' side effect

internal/controller/iprangeclaim_controller.go
```diff
@@ -0,0 +1,323 @@
...
+			ipRange.Spec.Description = updatedIpRangeSpec.Description
+			ipRange.Spec.PreserveInNetbox = updatedIpRangeSpec.PreserveInNetbox
+			return nil
+		})
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	if apismeta.IsStatusConditionTrue(ipRange.Status.Conditions, "Ready") {
```
alexandernorth, line 143: I would reverse the logic here, if we are not ready, end reconcile, if we are, then perform the status update logic (guard-like pattern)

alexandernorth (changes requested):
Nice work, please take a look at the comments and let me know if there is something I should clarify :)

api/v1/iprange_types.go
```diff
@@ -0,0 +1,113 @@
...
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:storageversion
+//+kubebuilder:printcolumn:name="StartAddress",type=string,JSONPath=`.spec.startAddress`
+//+kubebuilder:printcolumn:name="EndAddress",type=string,JSONPath=`.spec.endAddress`
+//+kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+//+kubebuilder:printcolumn:name="ID",type=string,JSONPath=`.status.id`
+//+kubebuilder:printcolumn:name="URL",type=string,JSONPath=`.status.url`
+//+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+// +kubebuilder:resource:shortName=ir
```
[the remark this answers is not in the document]
bruelea: I first used 'ipr' and 'iprc' as short names. But I'm so used to the claim CR only having 3 letters in the short name, that I have the impression that if we keep it the same here it would be more user friendly.

internal/controller/iprange_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const IpRangeFinalizerName = "iprange.netbox.dev/finalizer"
+const LastIpRangeMetadataAnnotationName = "iprange.netbox.dev/last-ip-range-metadata"
```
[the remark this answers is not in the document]
bruelea: initially the idea behind calling it ip-range-metadata was that we could add more data if a use case comes up. If we don't expect that the annotation name can be changed. Should it then also be changed for ip address and prefix?

internal/controller/iprange_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	Scheme            *runtime.Scheme
+	NetboxClient      *api.NetboxClient
+	Recorder          record.EventRecorder
+	OperatorNamespace string
+	RestConfig        *rest.Config
+}
+
+//+kubebuilder:rbac:groups=netbox.dev,resources=ipranges,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=netbox.dev,resources=ipranges/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=netbox.dev,resources=ipranges/finalizers,verbs=update
```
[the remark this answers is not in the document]
bruelea: the permission is added in another controller. Would it be good to have it here too for documentation?

internal/controller/iprange_controller.go
```diff
@@ -0,0 +1,306 @@
...
+
+		err := r.NetboxClient.DeleteIpRange(o.Status.IpRangeId)
+		if err != nil {
+			err = r.logErrorSetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpRangeReadyFalseDeletionFailed,
+				corev1.EventTypeWarning, "", err)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+
+			return ctrl.Result{Requeue: true}, nil
```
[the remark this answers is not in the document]
bruelea: The logger which is set up in cmd/main.go will log the stack trace for every error returned by a reconcile function. Because we don't need the stack trace for netbox api errors we log the error in the reconcile loop directly and return nil.

api/v1/iprangeclaim_types.go
```diff
@@ -0,0 +1,144 @@
...
+	Reason:  "IpRangeCRNotCreated",
+	Message: "Failed to fetch new Ip Range from NetBox",
+}
+
+var ConditionIpRangeAssignedFalseSizeMissmatch = metav1.Condition{
+	Type:    "IpRangeAssigned",
+	Status:  "False",
+	Reason:  "IpRangeCRNotCreated",
+	Message: "Assigned/Resored ip range has less available ip addresses than requested",
+}
```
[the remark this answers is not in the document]
henrybear327: I think we use IpAddress at other places already, so maybe `Ip` would be more consistent. :)

api/v1/iprangeclaim_types.go
```diff
@@ -0,0 +1,144 @@
...
+	Status:  "False",
+	Reason:  "IpRangeCRNotCreated",
+	Message: "Failed to fetch new Ip Range from NetBox",
+}
+
+var ConditionIpRangeAssignedFalseSizeMissmatch = metav1.Condition{
+	Type:    "IpRangeAssigned",
+	Status:  "False",
+	Reason:  "IpRangeCRNotCreated",
+	Message: "Assigned/Resored ip range has less available ip addresses than requested",
```
henrybear327, line 143:
```suggestion
	Message: "Assigned/Restored ip range has less available ip addresses than requested",
```

internal/controller/iprange_controller.go
```diff
@@ -0,0 +1,306 @@
...
+		if err := r.Update(ctx, o); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// 1. try to lock lease of parent prefix if IpRange status condition is not true
+	// and IpRange is owned by an IpRangeClaim
+	or := o.ObjectMeta.OwnerReferences
+	var ll *leaselocker.LeaseLocker
+	if len(or) > 0 /* len(nil array) = 0 */ && !apismeta.IsStatusConditionTrue(o.Status.Conditions, "Ready") {
```
henrybear327, line 120: Let's remove this comment `/* len(nil array) = 0 */`. It's something that I somehow left behind when fixing the linter error long ago... Sorry

```suggestion
	if len(or) > 0 && !apismeta.IsStatusConditionTrue(o.Status.Conditions, "Ready") {
```

internal/controller/iprange_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	if conditionChanged {
+		r.Recorder.Event(o, eventType, condition.Reason, condition.Message)
+	}
+
+	err := r.Client.Status().Update(ctx, o)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
henrybear327, line 249:
```suggestion
	return r.Client.Status().Update(ctx, o)
```

internal/controller/iprangeclaim_controller.go
```diff
@@ -0,0 +1,323 @@
...
+
+		// check if the restored ip range has the size requested by the claim
+		availableIpRanges, err := r.NetboxClient.GetAvailableIpAddressesByIpRange(ipRangeModel.Id)
+		if len(availableIpRanges.Payload) != o.Spec.Size {
+			ll.Unlock()
+			err = r.logErrorSetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpRangeAssignedFalseSizeMissmatch, corev1.EventTypeWarning, "", err)
+			if err != nil {
+				return nil, ctrl.Result{}, err
+			}
+			return nil, ctrl.Result{Requeue: true}, nil
```
henrybear327, line 318: Do we want to requeue in this case?

IMHO if `len(availableIpRanges.Payload) != o.Spec.Size` is triggered, it would require human intervention to fix it.

kind/load-data-job/local-demo-data.sql
```diff
@@ -17,6 +17,20 @@ VALUES (6, 'boolean', 'cfDataTypeBool', 'cf Data Type Bool', 'Custom field 3 for
 INSERT INTO public.extras_customfield (id, type, name, label, description, required, filter_logic, "default", weight, validation_minimum, validation_maximum, validation_regex, created, last_updated, related_object_type_id, group_name, search_weight, is_cloneable, choice_set_id, ui_editable, ui_visible, comments)
 VALUES (7, 'integer', 'cfDataTypeInteger', 'cf Data Type Integer', 'Custom field 4 for ParentPrefixSelector', false, 'exact', NULL, 100, NULL, NULL, '', '2024-06-13 15:17:08.65334+00', '2024-06-13 15:17:08.653359+00', NULL, 'netbox-operator', 100, false, NULL, 'hidden', 'always', '');
 
+-- associate custom fields IP Range
```
henrybear327, line 20:
```suggestion
-- associate custom fields with IP Range
```

kind/load-data-job/local-demo-data.sql
```diff
@@ -17,6 +17,20 @@ VALUES (6, 'boolean', 'cfDataTypeBool', 'cf Data Type Bool', 'Custom field 3 for
...
+INSERT INTO public.extras_customfield_object_types (id, customfield_id, objecttype_id)
+VALUES (6, 2, 78);
+
+INSERT INTO public.extras_customfield_object_types (id, customfield_id, objecttype_id)
+VALUES (7, 3, 78);
+
+-- associate custom fields IP Range
```
henrybear327, line 27:
```suggestion
-- associate custom fields with IP Range
```

henrybear327: Propose to move lines 21-25 to after `-- insert User Token`, keeping the entries in non-decreasing order

pkg/netbox/api/ip_range_claim.go
```diff
@@ -0,0 +1,190 @@
...
+	requestAvailableIPs := ipam.NewIpamIPRangesAvailableIpsListParams().WithID(ipRangeId)
+	responseAvailableIPs, err := r.Ipam.IpamIPRangesAvailableIpsList(requestAvailableIPs, nil)
+	if err != nil {
+		return nil, err
+	}
+	return responseAvailableIPs, nil
+}
+
+func searchAvailableIpRange(availableIps *ipam.IpamPrefixesAvailableIpsListOK, requiredSize int) (string, string, error) {
+	// this function receives a list of available IPs it chan have IPv4 or IPv6 IPs
```
henrybear327, line 120:
```suggestion
	// this function receives a list of available IPs that can be either IPv4 or IPv6 IPs
```

pkg/netbox/api/ip_range_claim.go
```diff
@@ -0,0 +1,190 @@
...
+func (r *NetboxClient) GetAvailableIpAddressesByIpRange(ipRangeId int64) (*ipam.IpamIPRangesAvailableIpsListOK, error) {
+	requestAvailableIPs := ipam.NewIpamIPRangesAvailableIpsListParams().WithID(ipRangeId)
+	responseAvailableIPs, err := r.Ipam.IpamIPRangesAvailableIpsList(requestAvailableIPs, nil)
+	if err != nil {
+		return nil, err
+	}
+	return responseAvailableIPs, nil
+}
+
+func searchAvailableIpRange(availableIps *ipam.IpamPrefixesAvailableIpsListOK, requiredSize int) (string, string, error) {
```
henrybear327, line 119: Just curious, if `availableIps` is guaranteed to be sorted.

henrybear327:
Good work @bruelea :)

api/v1/iprange_types.go
```diff
@@ -0,0 +1,113 @@
...
+	//+kubebuilder:validation:Format=cidr
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'startAddress' is immutable"
+	//+kubebuilder:validation:Required
+	StartAddress string `json:"startAddress"`
+
+	// the endAddress is the last ip address included in the ip range
+	//+kubebuilder:validation:Format=cidr
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'endAddress' is immutable"
+	//+kubebuilder:validation:Required
+	EndAddress string `json:"endAddress"`
```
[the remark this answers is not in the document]
bruelea: Let's add it when come across a simple method to validate this

api/v1/iprange_types.go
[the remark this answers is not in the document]
bruelea: changed both short names

api/v1/iprangeclaim_types.go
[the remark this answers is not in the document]
bruelea: I changed it to 'IP' in the Type, Status, Reason and Message fields, so it's consistent with the conditions of the other types. In the other parts of the code we use 'Ip' because it's easier to read.

internal/controller/iprangeclaim_controller.go
bruelea: switched to using update

internal/controller/iprangeclaim_controller.go
bruelea: Done

internal/controller/iprangeclaim_controller.go
bruelea: This case needs human intervention, but with exponential back off there won't be too many retries and the claim could get ready if the conflicting data in NetBox is cleaned up. :)

pkg/netbox/api/ip_range_claim.go
bruelea: The list returned when calling the /api/ipam/prefixes/<id>/available-ips/ is sorted. 
But I now also added sorting to this function in case this would change in future.

internal/controller/iprange_controller.go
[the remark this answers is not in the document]
alexandernorth: Setting resolved as discussed

internal/controller/utils.go
```diff
@@ -17,31 +17,55 @@ limitations under the License.
...
-func updateLastMetadataAnnotation(annotations map[string]string, cusotmFields map[string]string) (map[string]string, error) {
-	// update lastIpRangeMetadata annotation
-	if annotations == nil {
-		annotations = make(map[string]string)
+func generateLastMetadataAnnotation(cusotmFields map[string]string) (string, error) {
+	if cusotmFields == nil {
+		cusotmFields = make(map[string]string)
+	}
+
+	lastIpRangeMetadata, err := json.Marshal(cusotmFields)
```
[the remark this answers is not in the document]
alexandernorth: the other instances of the typo will need to be done manually

pkg/netbox/api/ip_range_claim.go
```diff
@@ -0,0 +1,190 @@
...
+	// this function receives a list of available IPs it chan have IPv4 or IPv6 IPs
+	// it will search for the first available range of IPs with the required size
+	// it will return the start and end IP of the range
+	var startAddress, endAddress string
+	consecutiveCount := 0
+
+	for i := 0; i < len(availableIps.Payload); i++ {
```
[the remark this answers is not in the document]
alexandernorth: As discussed, not needed for this PR

internal/controller/iprange_controller.go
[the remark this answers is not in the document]
alexandernorth: Closing as discussed

internal/controller/iprange_controller.go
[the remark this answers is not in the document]
alexandernorth: Closing as discussed

internal/controller/iprange_controller_test.go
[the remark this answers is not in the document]
alexandernorth: Closing as discussed

internal/controller/ipaddress_controller.go
```diff
@@ -45,7 +45,7 @@ import (
 )
 
 const IpAddressFinalizerName = "ipaddress.netbox.dev/finalizer"
-const LastIpAddressMetadataAnnotationName = "ipaddress.netbox.dev/last-ip-address-metadata"
+const LastIpAddressManagedCustomFieldsAnnotationName = "ipaddress.netbox.dev/managed-custom-fields"
```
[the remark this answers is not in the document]
bruelea: renamed it for the 3 controllers

internal/controller/iprangeclaim_controller.go
```diff
@@ -68,20 +70,32 @@ func (r *IpRangeClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			if !apierrors.IsNotFound(err) {
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{}, removeFinalizer(ctx, r.Client, o, IpRangeClaimFinalizerName)
+		}
+
+		err = r.Client.Delete(ctx, ipRange)
+		if !apierrors.IsNotFound(err) {
+			return ctrl.Result{}, err
+		}
```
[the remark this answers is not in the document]
bruelea: For garbage collection the deletion time stamp on the owned resource is only set after the owner got deleted. With the finalizer in the claim, the deletion needs to be handled in the reconcile loop.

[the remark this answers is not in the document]
alexandernorth: True, you are correct.

pkg/netbox/api/ip_range_claim.go
```diff
@@ -129,13 +129,9 @@ func searchAvailableIpRange(availableIps *ipam.IpamPrefixesAvailableIpsListOK, r
...
+	for i := 1; i < len(ips); i++ {
 		currentIp := ips[i]
-
-		var previousIP net.IP
-		if i > 0 {
-			previousIP = ips[i-1]
-		}
+		previousIP := ips[i-1]
 
 		if i == 0 || areConsecutiveIPs(previousIP, currentIp) {
```
alexandernorth, line 136: `i` will never be 0 now, this condition can be simplified

internal/controller/utils.go
```diff
@@ -17,31 +17,55 @@ limitations under the License.
...
 
 func convertCIDRToLeaseLockName(cidr string) string {
 	return strings.ReplaceAll(strings.ReplaceAll(cidr, "/", "-"), ":", "-")
 }
 
-func updateLastMetadataAnnotation(annotations map[string]string, cusotmFields map[string]string) (map[string]string, error) {
-	// update lastIpRangeMetadata annotation
-	if annotations == nil {
-		annotations = make(map[string]string)
+func generateLastMetadataAnnotation(cusotmFields map[string]string) (string, error) {
```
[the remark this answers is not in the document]
bruelea: Applied typo fixes

pkg/netbox/api/ip_range_claim.go
henrybear327: Cool, thanks!

henrybear327:
LGTM! When CI pipelines are fixed and the comments are all resolved, I will approve it ;) 

Good job @bruelea

internal/controller/utils.go
```diff
@@ -17,31 +17,55 @@ limitations under the License.
...
+	if cusotmFields == nil {
+		cusotmFields = make(map[string]string)
+	}
+
+	lastIpRangeMetadata, err := json.Marshal(cusotmFields)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal lastIpRangeMetadata annotation: %w", err)
```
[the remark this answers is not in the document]
bruelea: updated the error message

pkg/netbox/api/ip_range_claim.go
bruelea: removed this check

henrybear327:
Please create a tracking GitHub issue for the IPRange Testing :)

alexandernorth (approved):
Great work @bruelea !

api/v1/iprange_types.go
```diff
@@ -0,0 +1,113 @@
...
+
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
```
henrybear327, line 25:
```suggestion
// +kubebuilder:validation:XValidation:rule="self.startAddress < self.endAddress"
```

This is tested and worked in my local netbox. @alexandernorth @bruelea what do you think?

henrybear327 (approved):
Good job @bruelea !

api/v1/iprangeclaim_types.go
```diff
@@ -0,0 +1,144 @@
...
+type IpRangeClaimSpec struct {
+	//+kubebuilder:validation:Required
+	//+kubebuilder:validation:Format=cidr
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'parentPrefix' is immutable"
+	ParentPrefix string `json:"parentPrefix"`
+
+	//+kubebuilder:validation:Required
+	//+kubebuilder:validation:Minimum=2
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'size' is immutable"
+	Size int `json:"size,omitempty"`
```
jstudler, line 36: There is a current de-facto hard limit here which is at 50 (should be reproducible by trying to create a iprc with .spec.size 51 in an empty prefix) and a soft limit that might be even lower (depending on the fragmentation of IPs in the prefix). I assume this is due to pagination in the client. I propose to add validation Maximum=50 here and add a comment to this field.

api/v1/iprange_types.go
henrybear327: After discussion, we agree that for IPv6, it might not work due to the various formats you can represent it, so we will NOT implement this validation now

api/v1/iprangeclaim_types.go
bruelea: Added a validation rule for maximum value for size until pagination is added to the function searching an available iprange.

jstudler: Can you maybe make the comment a bit more verbose and more about the why (the what is covered in the validation). Something like:

"Size is the amount of consecutive IP Addresses you wish to reserve. Currently only sizes up to 50 are supported due to pagination of the NetBox API. In practice, this might be even lower depending on the fragmentation of the parent prefix."

bruelea: Updated the description with more information.

jstudler (approved):
let's go. thanks for the work!

PR #138, opened 2024-11-20
Add support for built-in field `family` in `ParentPrefixSelector`

henrybear327:
Added support for the built-in field `family` in `ParentPrefixSelector`.

The CR validation is also written to reject values besides `IPv4` and `IPv6`, when the key `family` is present in `ParentPrefixSelector`.

The documentation and the sample CR are also updated.

### DIFF
--- a/api/v1/prefixclaim_types.go
+++ b/api/v1/prefixclaim_types.go
@@ -39,9 +39,10 @@ type PrefixClaimSpec struct {
 	// The `parentPrefixSelector` is a key-value map, where all the entries are of data type `<string-string>`
 	// The map contains a set of query conditions for selecting a set of prefixes that can be used as the parent prefix
 	// The query conditions will be chained by the AND operator, and exact match of the keys and values will be performed
-	// 2 built-in fields, namely `tenant` and `site`, along with custom fields, can be used
+	// The built-in fields `tenant`, `site`, and `family`, along with custom fields, can be used
 	// For more information, please see ParentPrefixSelectorGuide.md
 	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'parentPrefixSelector' is immutable"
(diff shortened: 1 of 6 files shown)


api/v1/prefixclaim_types.go
```diff
@@ -39,9 +39,10 @@ type PrefixClaimSpec struct {
 	// The `parentPrefixSelector` is a key-value map, where all the entries are of data type `<string-string>`
 	// The map contains a set of query conditions for selecting a set of prefixes that can be used as the parent prefix
 	// The query conditions will be chained by the AND operator, and exact match of the keys and values will be performed
-	// 2 built-in fields, namely `tenant` and `site`, along with custom fields, can be used
+	// The built-in fields `tenant`, `site`, and `family`, along with custom fields, can be used
 	// For more information, please see ParentPrefixSelectorGuide.md
 	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'parentPrefixSelector' is immutable"
+	//+kubebuilder:validation:XValidation:rule="!has(self.family) || (self.family == 'IPv4' || self.family == 'IPv6')"
```
henrybear327, line 45: @jstudler I am not a network expert, please let me know if you think this might be too restrictive for validation. Thanks :)

jstudler: No this is fine. If specified, it has to be either IPv4 or IPv6. We could even think about removing `!has(self.family) ||  ` to force the user to specify the address family when using `parentPrefixSelector`. Otherwise, he might find a non-deterministic behaviour.

jstudler (approved):
lgtm. thanks @henrybear327 !

api/v1/prefixclaim_types.go
jstudler: We can leave this optional for now. The user has a way to prevent this so it should be fine.

henrybear327:
> lgtm. thanks @henrybear327 !

Thanks for reviewing @jstudler

PR #153, opened 2024-12-04
Add a prefix with IPv6 addresses to local demo data

henrybear327:
Alongside the IPv4 pools, we introduce IPv6 prefixes to the local demo data to make testing other operators easier.


jstudler (approved):
lgtm. I slightly updated the description and title to match with NetBox terminology.

henrybear327:
> lgtm. I slightly updated the description and title to match with NetBox terminology.

Thanks @jstudler !

PR #160, opened 2024-12-09
Cancel running workflow when new commits are pushed in a pull request

henrybear327:
Reduce noise in the case that CI jobs failed when new commits are pushed before the previous one's CI job has finished


alexandernorth (approved):
lgtm

PR #161, opened 2024-12-10
Add e2e tests for Prefix

henrybear327:
Leveraging Chainsaw, we can easily write e2e tests for the NetBox operator, testing directly against the NetBox instance

This PR focuses solely on the Prefix and PrefixClaim, happy path.

Implements https://github.com/netbox-community/netbox-operator/issues/163

### DIFF
--- a/.github/env
+++ b/.github/env
@@ -1,0 +1,3 @@
+golang-version=1.23
+kind-version=v0.25.0
+kind-image=kindest/node:v1.31.2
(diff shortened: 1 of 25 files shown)


henrybear327:
The happy path written are 
- apply the sample CR and see if it works
- apply -> delete -> apply and see if the restoration works

henrybear327:
The e2e test run is consistently running around 8.5 min (https://github.com/netbox-community/netbox-operator/actions/runs/12273670479/job/34244939819?pr=161)

The main time is spent on setting things up...

henrybear327:
@bruelea commented that we shouldn't load demo data for e2e tests, which I will address in https://github.com/netbox-community/netbox-operator/pull/164

henrybear327:
> @bruelea commented that we shouldn't load demo data for e2e tests

This is a lot more work than expected, since we need to extract the minimal tables to set up, etc. Propose to have it done in a follow-up PR. WDYT?

Thanks

bruelea (approved):
Nice work! Great to be able to test against an NetBox instance in the kind cluster.

PR #164, opened 2024-12-11
Upgrade NetBox image version

henrybear327:
There are several aspects to be updated:
- In the `local-env.sh` script, the images listed in `Images` would need to be changed, as updating the helm chart will cause the required image versions to change, and we need to pre-load them into the kind cluster since it can't pull images directly from the outside world (I checked `kubectl describe/logs` to see what images failed to load, and update the version strings accordingly)
- If the major/minor version is changed, we need to update the NetBox demo data link [1] in the `load-data.sh` script 

Based on the experience of migrating from 4.0.x to 4.1.x, the database schema might change. I dumped the database out by executing `kubectl exec pod/netbox-db-0 -- bash -c "pg_dump --clean -U postgres netbox" > database.sql` before and after the upgrade to compare the difference. 

Reference:
[1] https://github.com/netbox-community/netbox-demo-data/blob/master/sql/netbox-demo-v4.1.sql


alexandernorth (approved):
Looks good to me, thanks

PR #166, opened 2024-12-12
[IpRange] Add missing return after calling ReserveOrUpdateIpRange failed

henrybear327:
I [wrapped the error](https://go.dev/blog/go1.13-errors), so we will be able to see what happened before the logging failed.


alexandernorth (approved):
Thanks looks good to me

PR #167, opened 2024-12-12
Support e2e tests against multiple NetBox versions

henrybear327:
Since we are introducing the e2e test and running against real NetBox instances, it makes sense to test against different versions of NetBox, so we can flag issues early on.

CI will now run all e2e tests against the supported versions.

The supported NetBox versions are (the latest patch version for the `major.minor` release):
- 3.7.8
- 4.0.11
- 4.1.7 (4.1.8 not yet supported by the NetBox Helm chart yet)

Instead of `create-kind`, we now have `create-kind-3.7.8`, `create-kind-4.0.11`, and `create-kind-4.1.7`. The same goes for `make test-e2e`.

The patching mechanism is used to apply database schema fixes and demo data version changes. When we introduce support for new versions, usually only the database schema would need to be changed.


henrybear327:
There might be an impact on https://github.com/netbox-community/netbox-chart/pull/397 if the SQL data-loading mechanism is leveraged, as I might be splitting up the SQL file to support different NetBox versions.

Makefile
```diff
@@ -237,6 +237,26 @@ generate_mocks: ## TODO: auto install go install go.uber.org/mock/mockgen@latest
 	mkdir -p ${GEN_DIR}
 	mockgen -destination ${GEN_DIR}/${NETBOX_MOCKS_OUTPUT_FILE} -source=${INTERFACE_DEFITIONS_DIR}
 
-.PHONY: test-e2e
-test-e2e: install-$(GO_PACKAGE_NAME_CHAINSAW)
-	chainsaw test --namespace e2e
+# e2e tests
+# setting parallel to 2 to avoid flakiness
+E2E_PARAM := --namespace e2e --parallel 3 --apply-timeout 3m --assert-timeout 3m --delete-timeout 3m --error-timeout 3m --exec-timeout 3m --skip-delete
```
alexandernorth, line 242: Setting to 2 but actually set to 3 - which is it supposed to be?

kind/local-env.sh
alexandernorth, line 1: I wonder if we can change this to be a bit more generic. Right now, if we update the version in the job script, this file will then break. It also seems hard to maintain in the future when new versions are released, or we want to test against other versions too.

Maybe it would be worth converting this to use templating so that we can keep versioned versions of these - e.g. have one file per version containing the image dependencies, the Helm chart and anything else, and insert these into the script and job dynamically.

Happy to hear other ideas too.

alexandernorth (changes requested):

Makefile
henrybear327: I will remove the comment :D 
Good catch!

kind/local-env.sh
alexandernorth: Here are some options:
- [heredoc](https://tldp.org/LDP/abs/html/here-docs.html) (e.g. example 19.5) 
- [envsubst](https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html)

These would then use variables to template strings, this way we don't need to worry about modifying multiple files either.

henrybear327: Thanks for the pointers, @alexandernorth ! :) 

After looking into those 2 tools, I think we will more or less end up in the same scripting (`sed`) situation since the main issue lies in the database schema changes (e.g., column name changes, new columns added, etc.). The fields that we really care about, e.g., IP, Name, etc., are actually not changed at all. 

So I think in the end, maybe writing a small code-gen tool would be the most maintainable solution, e.g. define the IP, name, association with custom fields, etc., and by specifying the NetBox version we are targeting at, we generate the `database.sql` file accordingly.

henrybear327: Upon offline discussion with @jstudler, we would be implementing https://github.com/netbox-community/netbox-operator/issues/175 before merging this PR.

henrybear327:
Blocked by https://github.com/netbox-community/netbox-operator/issues/175

henrybear327:
Recreated at https://github.com/netbox-community/netbox-operator/pull/176

Issue #175, opened 2024-12-16
Migrate populating local demo data through database.sql to using NetBox API


henrybear327:
### What would you like to be added?

Migrate populating local demo data through `database.sql` to through using NetBox API

### Why is this needed?

When working on https://github.com/netbox-community/netbox-operator/pull/167, we discovered that it's hard to maintain multiple versions of the database schema + data for each supported NetBox versions. After revisiting the current approach we are using, we figured these local demo data should be handled through API calls, so we don't tamper directly with the database and it will work by default with multiple NetBox versions.

PR #176, opened 2024-12-16
Support e2e tests against multiple NetBox versions

henrybear327:
(Reopen https://github.com/netbox-community/netbox-operator/pull/167. Due to PR base branch setting errors, I can't reopen the original PR...)

Since we are introducing the e2e test and running against real NetBox instances, it makes sense to test against different versions of NetBox, so we can flag issues early on.

CI will now run all e2e tests against the supported versions.

The supported NetBox versions are (the latest patch version for the `major.minor` release):
- 3.7.8
- 4.0.11
- 4.1.8

Instead of `create-kind`, we now have `create-kind-3.7.8`, `create-kind-4.0.11`, and `create-kind-4.1.8`. The same goes for `make test-e2e`.

The patching mechanism is used to apply database schema fixes and demo data version changes. When we introduce support for new versions, usually only the database schema would need to be changed.

Blocked by https://github.com/netbox-community/netbox-operator/issues/175 (https://github.com/netbox-community/netbox-operator/pull/177)

### DIFF
--- a/kind/load-data-job/main.py
+++ b/kind/load-data-job/main.py
@@ -1,7 +1,6 @@
 import pynetbox
 from pprint import pprint
 from dataclasses import dataclass
-import sys
 
 print("Starting to load data onto NetBox through API")
 try:
@@ -11,7 +10,7 @@ try:
     )
 except pynetbox.RequestError as e:
     pprint(e.error)
-    sys.exit(1)
+   
 print("Connected to NetBoxAPI")
 
 # insert Tenants
@@ -47,7 +46,7 @@ for tenant in tenants:
         )
     except pynetbox.RequestError as e:
         pprint(e.error)
-        sys.exit(1)
+       
 print("Tenants loaded")
 
 # insert Sites
@@ -88,13 +87,14 @@ for site in sites:
         )
     except pynetbox.RequestError as e:
         pprint(e.error)
-        sys.exit(1)
+       
 print("Sites loaded")
 
 # create custom fields and associate custom fields with IP/IPRange/Prefix
 @dataclass
 class CustomField:
-    object_types: list[str]
+    content_types: list[str] # for v3
+    object_types: list[str] # for v4
     type: str
     name: str
     label: str
@@ -104,6 +104,7 @@ class CustomField:
 
 custom_fields = [
     CustomField(
+        content_types=["ipam.ipaddress", "ipam.iprange", "ipam.prefix"],
         object_types=["ipam.ipaddress", "ipam.iprange", "ipam.prefix"],
         type="text",
         name="netboxOperatorRestorationHash",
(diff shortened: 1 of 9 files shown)


henrybear327:
Still fixing 3.7.8 support. Something is failing somehow...

jstudler (approved):

henrybear327:
To test this locally, execute `make test-e2e-4.1.8` (or other supported versions)

kind/local-env.sh
```diff
@@ -1,44 +1,118 @@
...
+  exit 1
 fi
 
+for img in "${Remote_Images[@]}"; do
+  docker pull "$img"
+  kind load docker-image "$img"
+done
+
 # build image for loading local data via NetBox API
 cd ./kind/load-data-job && docker build -t netbox-load-local-data:1.0 --no-cache --progress=plain -f ./dockerfile . && cd -
```
jstudler, line 106: On my macOS Podman 5.2.4, this will build but not tag the image when running `make create-kind`:

```bash
...
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load

What's Next?
  View summary of image vulnerabilities and recommendations → docker scout quickview
/tmp/netbox-operator
using podman due to KIND_EXPERIMENTAL_PROVIDER
enabling experimental podman provider
ERROR: image: "netbox-load-local-data:1.0" not present locally
make: *** [create-kind] Error 1
```

`docker image ls` will not show the image then. If I swap from the `docker` cli to the `podman` cli, it works.

I think adding the `--load` parameter should fix this for podman:

```suggestion
cd ./kind/load-data-job && docker build -t netbox-load-local-data:1.0 --load --no-cache --progress=plain -f ./dockerfile . && cd -
```

henrybear327: Testing CI now

kind/load-data-job/main.py
jstudler, line 1: Could we make this idempotent so it doesn't fail when executed multiple times? 

Currently it will fail when an object already exists. Example:

```
('{"name":["tenant with this name already exists."],"slug":["tenant with this '
 'slug already exists."]}')
```

This means when the job exits e.g. because there was a communication hiccup between the job pod and NetBox API it will never be able to succeed after.

henrybear327: Sure! I will remove the assertion!

I intentionally designed it to fail in case I did something wrong to the script. But seems like there is a bigger issue with job execution :(

kind/local-env.sh
henrybear327: Passed.

kind/load-data-job/main.py
henrybear327: Done as requested.

henrybear327:
@alexandernorth PTAL thank you :)

alexandernorth (approved):
I think this is ok to add, perhaps we should revisit the best way to handle multiple version services in the future

PR #177, opened 2024-12-16
Migrate populating local demo data through database.sql to using NetBox API 

henrybear327:
Several changes are introduced in this PR
- `database.sql` is stripped down to only populating the API token
- The local demo data is now populated through NetBox API, by using a Python script and running as a Kubernetes job, which makes it easier to work across NetBox versions. (idea from @jstudler)

e2e tests are modified accordingly, as we also migrate to using locally-created tenants and sites only (idea from @bruelea ). We are also check if the tenant and site from parentPrefixSelector and the spec is properly leveraged by using 2 different tenants and sites.

Implements and closes https://github.com/netbox-community/netbox-operator/issues/175


henrybear327:
There are 2 `TODO(henrybear327)` that will be deferred for the future, as it's NetBox implementation related issues

henrybear327:
To run the e2e test locally, please use a clean kind cluster (you can remove the current one by executing `kind delete cluster --name kind`), and execute `make create-kind deploy-kind test-e2e` @faebr

faebr (approved):
Looks way clearer now with the python script I like it. Also tested it locally on my machine and works as expected.

PR #221, opened 2025-01-27
[CI] Do not cancel concurrent CI jobs on main

henrybear327:
When we merge a PR into the main branch before the previously merged commit finishes, the CI jobs will be cancelled (see https://github.com/netbox-community/netbox-operator/commits/main/), which is not ideal as we would like to see if all commits merged into the main branch can pass CI checks.

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -9,7 +9,7 @@ on:
   pull_request:
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ !contains(github.ref, 'main/')}}
 env:
   IMAGE_NAME: netbox-operator
   DOCKER_METADATA_PR_HEAD_SHA: true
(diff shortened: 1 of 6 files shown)


.github/workflows/build-image.yaml
```diff
@@ -9,7 +9,7 @@ on:
   pull_request:
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ !contains(github.ref, 'main/')}}
```
henrybear327, line 12: The same should be added for v* in the case where we do releases in quick succession.

henrybear327:
Hey team, 

Could we have this reviewed and merged when time permits :) 

This will reduce the CI failed email bombing when we merge dependabot PRs in quick succession! Thanks

bruelea (approved):
Agree! Jobs shouldn't be canceled on the main branch. Thanks for implementing it.

henrybear327:
This is somehow not working on main branch. I will fix this in a follow-up PR...

PR #257, opened 2025-03-06
Remove duplicate kind cluster creation

jstudler:
Kind clusters for the e2e tests are implicitly created in the `make test-e2e-*` targets that are triggered in the `Run e2e tests` steps of the `e2e-tests` pipeline. Thus, there is no need for the additional `Start kind cluster` step in the e2e tests pipeline. This PR removes the additional kind cluster creation.


henrybear327:
If the test can pass, we can indeed remove those setup parts :)
Good catch!

henrybear327 (approved):

PR #263, opened 2025-03-10
Improve CRD documention

### LINKED ISSUE #128: Improve CRD documention
### What would you like to be added?

Proper documentation for each of the fields in all of the CRD.

### Why is this needed?

Let's take the documentation for PrefixClaim as an example. 

As we can see, a lot of fields have `<no description>` as it's documentation, which is not ideal.

```bash
➜  netbox-operator-henry git:(feat/issue_79) ✗ kubectl explain prefixclaim.spec
GROUP:      netbox.dev
KIND:       PrefixClaim
VERSION:    v1

FIELD: spec <Object>


DESCRIPTION:
    PrefixClaimSpec defines the desired state of PrefixClaim
    TODO: The reason for using a workaround please see
    https://github.com/netbox-community/netbox-operator/pull/90#issuecomment-2402112475
    
FIELDS:
  comments      <string>
    <no description>

  customFields  <map[string]string>
    <no description>

  description   <string>
    <no description>

  parentPrefix  <string>
    <no description>

  parentPrefixSelector  <map[string]string>
    The `parentPrefixSelector` is a key-value map, where all the entries are of
    data type `<string-string>`.
    The map contains a set of query conditions for selecting a set of prefixes
    that can be used as the parent prefix.
    The query conditions will be chained by the AND operator, and exact match of
    the keys and values will be performed.

  prefixLength  <string> -required-
    <no description>

  preserveInNetbox      <boolean>
    <no description>

  site  <string>
    Use the `name` value instead of the `slug` value

  tenant        <string>
    Use the `name` value instead of the `slug` value
```

jstudler:
- Add more verbose CRD documentation
- Add Installation methods
- Add notes on Mixed usage of Prefixes
- Add FAQ

Fixes #128

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,15 @@ NetBox Operator extends the Kubernetes API by allowing users to manage NetBox re
 - kind
 - docker cli
 
+## Installation methods
+
+There are several ways to install NetBox Operator on your Cluster:
+
+- NetBox Operator Helm Chart. More information: <https://github.com/netbox-community/netbox-chart/>
+- Manifests in <config/default> of this repo. Make sure to first set the correct image using `kustomize edit set image controller=<correctimage>`. Images can be found on the Releases page on GitHub
+- The Makefile targets in this repository.
+- For debugging and developing, please read further below.
+
 # How to use NetBox Operator
 
 ## Running both NetBox Operator and NetBox on a local kind cluster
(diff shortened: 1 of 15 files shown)


pablogarciamiranda:
I would also suggest adding some documentation about required/optional fields, immutable fields, and default values, as this information can only be get by actually applying CRs.

pablogarciamiranda:
In addition to my other suggestion, I would also provide example values for some fields. This can be omitted if we have samples already containing all the necessary fields

pablogarciamiranda (approved):

PR #265, opened 2025-03-14
Consolidate Status and Condition reporting

alexandernorth:
This PR shows one way in which we can consolidate the reporting of status and conditions into a common function.

It exposes access to the EventRecorder so events can be emitted independently of condition updates, and also logs changes in conditions as well as errors.

Would be happy to receive feedback and take suggestions as to how we can improve this functionality.

### DIFF
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -89,7 +89,7 @@ func (esr *EventStatusRecorder) Report(ctx context.Context, o ObjectWithConditio
 	logger := log.FromContext(ctx)
 
 	if errExt != nil {
-		condition.Message = condition.Message + ", " + errExt.Error()
+		condition.Message = condition.Message + ": " + errExt.Error()
 		logger.Error(errExt, condition.Message)
 	}
 
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -22,6 +22,9 @@ import (
 	"fmt"
 	"strings"
 
+	apismeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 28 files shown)


internal/controller/utils.go
```diff
@@ -69,3 +72,41 @@ func addFinalizer(ctx context.Context, c client.Client, o client.Object, finaliz
...
+	if conditionChanged {
+		esr.rec.Event(o, eventType, condition.Reason, condition.Message)
+	}
+
+	err := esr.client.Status().Update(ctx, o)
+	if err != nil {
+		return err
+	}
+
+	logger.Info("Condition "+condition.Type+" changed to "+string(condition.Status), "Reason", condition.Reason, "Message", condition.Message)
```
bruelea, line 109: Are there cases where the Report method is called when the message logged here is different to the previous logs but the condition doesn't changes? Otherwise this log could move to line 102.

internal/controller/utils.go
```diff
@@ -69,3 +72,41 @@ func addFinalizer(ctx context.Context, c client.Client, o client.Object, finaliz
...
+}
+
+func NewEventStatusRecorder(client client.Client, rec record.EventRecorder) *EventStatusRecorder {
+	return &EventStatusRecorder{
+		client: client,
+		rec:    rec,
+	}
+}
+
+func (esr *EventStatusRecorder) Report(ctx context.Context, o ObjectWithConditions, condition metav1.Condition, eventType string, errExt error, additionalMessages ...string) error {
```
bruelea, line 88: For the Claim CRs would it be helpful if there's a check if the 'Ready' condition should be set to true/false based on the other conditions (for example IpAssigned) or should the claim_controller handle this? If a check should be used then a third condition like 'ChildResourceReady' would be needed.

bruelea (commented):
Nice, the possibility to call the Report method on the EventStatusRecorder makes it well readable, also the Conditions() method in the api is very clear.

internal/controller/utils.go
alexandernorth: I suppose this is to log whenever there is a desire to change the condition even if the condition itself has not changed. But then it should be higher up in the function anyway. Does this bring value or do we only want to log if there was a change to the condition?

internal/controller/utils.go
alexandernorth: My first thoughts are that this is controller-specific behaviour, and should probably be handled by a separate, specific function in the controller.
This function can use call this Report func, but it should be its own responsibility to know whether or not all other conditions are in a ready state

internal/controller/utils.go
bruelea: If the conditions gets status 'true' we don't need the log every reconcile loop. If it's false in most cases it would be because of an error, which is already logged earlier in the reconcile loop or if the child resource is not ready. For the case where the child resource is not ready the log in this line could be made with the normal logger https://github.com/netbox-community/netbox-operator/blob/57c7e613381a4eb29b1372025ede90915dc4c820/internal/controller/ipaddressclaim_controller.go#L210 or an error could be defined for this case, so that the EventStatusRecorder would create a log if the owned resource is not ready.

alexandernorth: Fair point - I have implemented this now

faebr (changes requested):
I very much like the usage of an interface and removing all this duplicate code. I have 2 improvement ideas, see comments.

internal/controller/utils.go
```diff
@@ -69,3 +72,40 @@ func addFinalizer(ctx context.Context, c client.Client, o client.Object, finaliz
...
+	condition.Message = strings.Join(append([]string{condition.Message}, additionalMessages...), ", ")
+	condition.ObservedGeneration = o.GetGeneration()
+
+	conditionChanged := apismeta.SetStatusCondition(o.Conditions(), condition)
+	if conditionChanged {
+		esr.rec.Event(o, eventType, condition.Reason, condition.Message)
+		logger.Info("Condition "+condition.Type+" changed to "+string(condition.Status), "Reason", condition.Reason, "Message", condition.Message)
+	}
+
+	err := esr.client.Status().Update(ctx, o)
```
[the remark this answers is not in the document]
alexandernorth: Done, conditionChanged also reports if ObservedGeneration has changed so this is also covered

bruelea (changes requested):

bruelea (approved):

faebr (approved):

PR #266, opened 2025-03-14
Check restoration hash in NetBox before updating resource in NetBox

bruelea:
In case the lock times out before a resource claimed by a prefix was successfully created in NetBox, two claims can get the same value assigned. By checking the restoration hash a resource has in NetBox assignments of the same resource to different claims can be avoided.

### DIFF
--- a/internal/controller/ipaddress_controller.go
+++ b/internal/controller/ipaddress_controller.go
@@ -170,10 +170,25 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	netboxIpAddressModel, err := r.NetboxClient.ReserveOrUpdateIpAddress(ipAddressModel)
 	if err != nil {
-		updateStatusErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpaddressReadyFalse,
-			corev1.EventTypeWarning, o.Spec.IpAddress)
-		return ctrl.Result{}, fmt.Errorf("failed to update ip address status: %w, "+
-			"after reservation of ip in netbox failed: %w", updateStatusErr, err)
+		if errors.Is(err, api.ErrRestorationHashMissmatch) && o.Status.IpAddressId == 0 {
+			// if there is a restoration has missmatch and the IpAddressId status field is not set,
+			// delete the ip address so it can be recreated by the ip address claim controller
+			logger.Info("restoration hash missmatch, deleting ip address custom resource", "ipaddress", o.Spec.IpAddress)
+			err = r.Client.Delete(ctx, o)
+			if err != nil {
+				updateStatusErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpaddressReadyFalse,
+					corev1.EventTypeWarning, err.Error())
+				return ctrl.Result{}, fmt.Errorf("failed to update ip address status: %w, "+
+					"after deletion of ip address cr failed: %w", updateStatusErr, err)
+			}
+			return ctrl.Result{}, nil
+		}
+		if updateStatusErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpaddressReadyFalse,
+			corev1.EventTypeWarning, o.Spec.IpAddress); updateStatusErr != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update ip address status: %w, "+
+				"after reservation of ip in netbox failed: %w", updateStatusErr, err)
+		}
+		return ctrl.Result{}, nil
 	}
 
 	// 3. unlock lease of parent prefix
--- a/internal/controller/ipaddress_controller.go
+++ b/internal/controller/ipaddress_controller.go
@@ -170,30 +170,22 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	netboxIpAddressModel, err := r.NetboxClient.ReserveOrUpdateIpAddress(ipAddressModel)
 	if err != nil {
-		if errors.Is(err, api.ErrRestorationHashMismatch) {
-			if o.Status.IpAddressId == 0 {
-				// if there is a restoration hash mismatch and the IpAddressId status field is not set,
-				// delete the ip address so it can be recreated by the ip address claim controller
-				// this will only affect resources that are created by a claim controller (and have a restoration hash custom field
-				logger.Info("restoration hash mismatch, deleting ip address custom resource", "ipaddress", o.Spec.IpAddress)
-				err = r.Client.Delete(ctx, o)
-				if err != nil {
-					if updateStatusErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpaddressReadyFalse,
-						corev1.EventTypeWarning, err.Error()); updateStatusErr != nil {
-						return ctrl.Result{}, fmt.Errorf("failed to update ip address status: %w, "+
-							"after deletion of ip address cr failed: %w", updateStatusErr, err)
-					}
-					return ctrl.Result{Requeue: true}, nil
+		if errors.Is(err, api.ErrRestorationHashMismatch) && o.Status.IpAddressId == 0 {
+			// if there is a restoration hash mismatch and the IpAddressId status field is not set,
+			// delete the ip address so it can be recreated by the ip address claim controller
+			// this will only affect resources that are created by a claim controller (and have a restoration hash custom field
+			logger.Info("restoration hash mismatch, deleting ip address custom resource", "ipaddress", o.Spec.IpAddress)
+			err = r.Client.Delete(ctx, o)
+			if err != nil {
+				if updateStatusErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpaddressReadyFalse,
+					corev1.EventTypeWarning, err.Error()); updateStatusErr != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to update ip address status: %w, "+
+						"after deletion of ip address cr failed: %w", updateStatusErr, err)
 				}
-				return ctrl.Result{}, nil
+				return ctrl.Result{Requeue: true}, nil
 			}
-		} else {
-			if updateStatusErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpaddressReadyFalse,
-				corev1.EventTypeWarning, err.Error()); updateStatusErr != nil {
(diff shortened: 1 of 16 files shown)


internal/controller/ipaddress_controller.go
```diff
@@ -170,10 +170,25 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	netboxIpAddressModel, err := r.NetboxClient.ReserveOrUpdateIpAddress(ipAddressModel)
 	if err != nil {
-		updateStatusErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpaddressReadyFalse,
-			corev1.EventTypeWarning, o.Spec.IpAddress)
-		return ctrl.Result{}, fmt.Errorf("failed to update ip address status: %w, "+
-			"after reservation of ip in netbox failed: %w", updateStatusErr, err)
+		if errors.Is(err, api.ErrRestorationHashMissmatch) && o.Status.IpAddressId == 0 {
+			// if there is a restoration has missmatch and the IpAddressId status field is not set,
+			// delete the ip address so it can be recreated by the ip address claim controller
```
jstudler, line 175: Please add a note here that this only applies if there is a restoration hash assigned to the resource which is only the case when the Claim Controller created the resource. Something like:

```
			// if there is a restoration hash mismatch and the IpAddressId status field is not set,
			// delete the ip address so it can be recreated by the ip address claim controller
			// this will only affect resources that are created by a claim controller (and have a restoration hash custom field)
```

jstudler (changes requested):
Looks good to me, would be great if @alexandernorth could also spend a few minutes to review the code.

A few minor things:
- "mismatch" is misspelled. search/replace "issmatch" with "ismatch"
- "restoration has" should be "restoration hash"

internal/controller/ipaddress_controller.go
```diff
@@ -170,10 +170,25 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			logger.Info("restoration hash missmatch, deleting ip address custom resource", "ipaddress", o.Spec.IpAddress)
+			err = r.Client.Delete(ctx, o)
+			if err != nil {
+				updateStatusErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpaddressReadyFalse,
+					corev1.EventTypeWarning, err.Error())
+				return ctrl.Result{}, fmt.Errorf("failed to update ip address status: %w, "+
```
alexandernorth, line 181: This is missing a check if `updateStatusErr != nil`

internal/controller/ipaddress_controller.go
```diff
@@ -170,10 +170,25 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+					"after deletion of ip address cr failed: %w", updateStatusErr, err)
+			}
+			return ctrl.Result{}, nil
+		}
+		if updateStatusErr := r.SetConditionAndCreateEvent(ctx, o, netboxv1.ConditionIpaddressReadyFalse,
+			corev1.EventTypeWarning, o.Spec.IpAddress); updateStatusErr != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update ip address status: %w, "+
```
alexandernorth, line 188: This is missing a check if `updateStatusErr != nil`

internal/controller/prefix_controller.go
```diff
@@ -184,8 +184,25 @@ func (r *PrefixReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-		return ctrl.Result{}, fmt.Errorf("failed at update prefix status: %w, "+"after reservation of prefix in netbox failed: %w", updateStatusErr, err)
+		if errors.Is(err, api.ErrRestorationHashMissmatch) && prefix.Status.PrefixId == 0 {
+			// if there is a restoration has missmatch and the PrefixId status field is not set,
+			// delete the prefix so it can be recreated by the prefix claim controller
+			logger.Info("restoration hash missmatch, deleting prefix custom resource", "prefix", prefix.Spec.Prefix)
+			err = r.Client.Delete(ctx, prefix)
+			if err != nil {
+				updateStatusErr := r.SetConditionAndCreateEvent(ctx, prefix, netboxv1.ConditionIpaddressReadyFalse,
+					corev1.EventTypeWarning, err.Error())
+				return ctrl.Result{}, fmt.Errorf("failed to update prefix status: %w, "+
```
alexandernorth, line 195: This is missing a check if `updateStatusErr != nil`

internal/controller/prefix_controller.go
```diff
@@ -184,8 +184,25 @@ func (r *PrefixReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+					"after deletion of prefix cr failed: %w", updateStatusErr, err)
+			}
+			return ctrl.Result{}, nil
+		}
+		if updateStatusErr := r.SetConditionAndCreateEvent(ctx, prefix, netboxv1.ConditionIpaddressReadyFalse,
+			corev1.EventTypeWarning, prefix.Spec.Prefix); updateStatusErr != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update prefix status: %w, "+
```
alexandernorth, line 202: This is missing a check if `updateStatusErr != nil`

pkg/netbox/api/ip_address_test.go
```diff
@@ -254,4 +275,89 @@ func TestIPAddress(t *testing.T) {
...
+
+		mockIPAddress.EXPECT().IpamIPAddressesList(inputList, nil).Return(outputList, nil).AnyTimes()
+		// use gomock.Any() because the input contains a pointer
+		mockIPAddress.EXPECT().IpamIPAddressesUpdate(gomock.Any(), nil).Return(outputUpdate, nil)
+
+		client := &NetboxClient{
+			Ipam: mockIPAddress,
+		}
+
+		ipAddressModel := ipAddressModel("fioaf9289rjfhaeuih")
```
alexandernorth, line 334: Should this be declared as a const - it appears elsewhere too

pkg/netbox/api/ip_range_test.go
```diff
@@ -183,6 +184,53 @@ func TestIpRange(t *testing.T) {
...
+		// ip range mock output
+		listOutput := &ipam.IpamIPRangesListOK{
+			Payload: &ipam.IpamIPRangesListOKBody{
+				Results: []*netboxModels.IPRange{
+					{
+						ID:           expectedIPRange().ID,
+						StartAddress: &startAddress,
+						EndAddress:   &endAddress,
+						CustomFields: map[string]interface{}{
+							config.GetOperatorConfig().NetboxRestorationHashFieldName: "different hash",
```
alexandernorth, line 203: Other controller uses something that looks like a hash - this could be aligned to either be text description or a hash for all controllers?

internal/controller/ipaddress_controller.go
bruelea: Added

pkg/netbox/api/ip_address_test.go
bruelea: I've now defined variable for all hashes, so it's also easier to read.

pkg/netbox/api/ip_range_test.go
bruelea: Agree, I chose to use sample hashes in all the tests.

bruelea:
> Looks good to me, would be great if @alexandernorth could also spend a few minutes to review the code.
> 
> A few minor things:
> 
> * "mismatch" is misspelled. search/replace "issmatch" with "ismatch"
> * "restoration has" should be "restoration hash"

Thanks, the typos are fixed now.

internal/controller/ipaddress_controller_test.go
```diff
@@ -81,31 +83,40 @@ var _ = Describe("IpAddress Controller", Ordered, func() {
...
-		}, timeout, interval).Should(BeTrue())
+
+		if restorationHashMismatch {
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: cr.GetName(), Namespace: cr.GetNamespace()}, createdCR)
+				return apierrors.IsNotFound(err)
+			}, timeout, interval).Should(BeTrue())
+		} else {
+
+			// check that reconcile loop did run a least once by checking that conditions are set
```
henrybear327, line 95:
```suggestion
			// check that reconcile loop did run at least once by checking that conditions are set
```

pkg/netbox/api/ip_address.go
```diff
@@ -52,7 +58,22 @@ func (r *NetboxClient) ReserveOrUpdateIpAddress(ipAddress *models.IPAddress) (*n
...
+
+	ipToUpdate := responseIpAddress.Payload.Results[0]
+
+	// if the desired ip address has a restoration hash
+	// check that the ip address to update has the same restoration hash
+	restorationHashKey := config.GetOperatorConfig().NetboxRestorationHashFieldName
+	if ipAddress.Metadata != nil {
+		if restorationHash, ok := ipAddress.Metadata.Custom[restorationHashKey]; ok {
+			if ipToUpdate.CustomFields != nil && ipToUpdate.CustomFields.(map[string]interface{})[restorationHashKey] == restorationHash {
+				//update ip address since it does exist and the restoration hash matches
```
henrybear327, line 70:
```suggestion
				// update ip address since it does exist and the restoration hash matches
```

pkg/netbox/api/ip_range.go
```diff
@@ -53,6 +61,22 @@ func (r *NetboxClient) ReserveOrUpdateIpRange(ipRange *models.IpRange) (*netboxM
...
+
+	ipRangeToUpdate := responseIpRange.Payload.Results[0]
+
+	// if the desired ip address has a restoration hash
+	// check that the ip address to update has the same restoration hash
+	restorationHashKey := config.GetOperatorConfig().NetboxRestorationHashFieldName
+	if ipRange.Metadata != nil {
+		if restorationHash, ok := ipRange.Metadata.Custom[restorationHashKey]; ok {
+			if ipRangeToUpdate.CustomFields != nil && ipRangeToUpdate.CustomFields.(map[string]interface{})[restorationHashKey] == restorationHash {
+				//update ip address since it does exist and the restoration hash matches
```
henrybear327, line 73:
```suggestion
				// update ip address since it does exist and the restoration hash matches
```

henrybear327 (changes requested):

internal/controller/ipaddress_controller_test.go
bruelea: Thanks, @henrybear327!

PR #286, opened 2025-04-16
Basic operational manual

pablogarciamiranda:
Basic operational manual for troubleshooting the NetBox Operator. The goal is to provide a quick-start guide for users and maintainers when diagnosing issues with the operator.

As soon as "common errors" are discovered, we can add them to this manual.


bruelea (changes requested):
Thanks! I like the collection of troubleshooting commands in the operational manual.

docs/claim-model.svg
faebr, line 1: I don't understand this graph, or it does not make sense to me. Why are there 2 netbox-operator boxes? How is this decision thing in the middle not in nextbox-operator and not in netbox itself? I would prefer a graph where it is visible that the prefix is written back to the prefixclaim. Also it should be visible what the user inputs. I would even do 2 different graphs for prefix and ipAddressClaim. Then have the user create the cr, netbox-operator creating the prefix, making the api calls and then writing the result to the cr. Or what was the goal of this graph? Maybe I also misunderstand?

docs/operational-manual.md
```diff
@@ -0,0 +1,43 @@
...
+
+```bash
+kubectl logs -n <namespace> deployment/netbox-operator-controller-manager
+```
+
+### Check CR Status
+Inspect the CRs status:
+
+```bash 
+kubectl describe <netbox-crd> <netbox-cr> -n <namespace>
```
faebr, line 19: It would be great to add an example here

faebr (changes requested):

docs/operational-manual.md
pablogarciamiranda: Examples added!

docs/claim-model.svg
pablogarciamiranda: You're right... what you're describing makes a lot more sense.

Actually, it would be more or less the same as this, but with the additional step of writing the actual prefix into the status of the PrefixClaim. Do you think it would make sense to duplicate this and include that?

https://github.com/netbox-community/netbox-operator/blob/main/docs/netbox-operator-high-level-architecture.drawio.svg

bruelea: This picture would work well. If you want you can also update the original picture you liked and reference it in both places instead of duplicating it or restructure the README in a way that the picture is only required once.

faebr: Yes I like that picture better. But I would remove the top example with the user writing the prefix driectly and only leave the bottom example using the claim. And yes good idea I would also add the prefix in the status. And then the same image could be used for ipAddress. I think the example with ipAddress is even easier to understand than with prefix. And yes agree to use the updated version from the README and from here

pablogarciamiranda: Ok, I will take the image and modify it slightly to include the write in the status and remove the part where the user writes.

I also think that only one, either PrefixClaim or IpAddressClaim, would be enough.

Regarding the updated version of the README, I’m not quite sure what you mean. Could you clarify?

pablogarciamiranda: Let me know what you think now @faebr @bruelea

faebr (changes requested):

docs/claim-model.svg
pablogarciamiranda: Let me know what you think now :)

I've exported the .SVG with "Appearance" = "Light" and seems to be the same as before even in GitHub dark mode.

pablogarciamiranda: > nice looks good! Since you have the workflow right now,, would you mind also updating the middle image with this Appearance so it also looks the same in dark mode?

Done ✅ 

> I like the picture :) What do you think about moving the new chapter about the claim model to the top of the README so that we do not need the same image twice in the RADME docs/netbox-operator-high-level-architecture.drawio.svg?

Done ✅

faebr (approved):

bruelea (approved):

PR #295, opened 2025-05-02
Feature/check keys in parent prefix selector

bruelea:
With this feature a check is added to only allow existing fields in the parentPrefixSelector.
If the parent prefix selector contains a key for a custom field that does not exist in NetBox, the key is ignored in the filter.

### DIFF
--- a/pkg/netbox/api/prefix_claim.go
+++ b/pkg/netbox/api/prefix_claim.go
@@ -195,7 +195,6 @@ func (r *NetboxClient) customFieldsExistsOrErr(customfieldFilterEntries []Custom
 
 	missingCustomFields := make([]string, 0)
 	for _, entry := range customfieldFilterEntries {
-		//request to netbox to check if the custom field existse
 		if !slices.Contains(customFieldNames, entry.key) {
 			missingCustomFields = append(missingCustomFields, entry.key)
 		}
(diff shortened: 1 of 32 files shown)


pkg/netbox/api/prefix_claim.go
```diff
@@ -158,6 +168,46 @@ func (r *NetboxClient) GetAvailablePrefixByParentPrefixSelector(prefixClaimSpec
...
+	missingCustomFields := make([]string, 0)
+	for _, entry := range customfieldFilterEntries {
+		//request to netbox to check if the custom field existse
+		if !slices.Contains(customFieldNames, entry.key) {
+			missingCustomFields = append(missingCustomFields, entry.key)
+		}
+	}
+
+	if len(missingCustomFields) > 0 {
+		return fmt.Errorf("invalid parentPrefixSelector, netbox custom fields %v do not exist", missingCustomFields)
```
faebr, line 205:
```suggestion
		return fmt.Errorf(
		    "invalid parentPrefixSelector, netbox custom fields %s do not exist",
		    strings.Join(missingCustomFields, ", "),
		)
```

faebr (changes requested):
Testes locally and works nice! I like the new thorough tests and the nice error message. Just 2 minor changes

faebr (approved):

tests/e2e/Prefix/IPv4/prefixclaim-ipv4-invalid-parentprefixselector/chainsaw-test.yaml
```diff
@@ -53,8 +53,5 @@ spec:
       description: Events cleanup required to fix issues with failing tests that assert the wrong Error resource
       cleanup:
         - script:
-            env:
-              - name: NAMESPACE
```
faebr: what is the reason for removing namespace here? just not needed since its already in the pipeline?

bruelea: It's not needed, looks like chainsaw defines this environment variable by default.
I thought it would simplify the tests if we remove it as it is redundant.

PR #335, opened 2025-06-23
Configure dependabot to run on mondays at 04:00 in the Europe/Zurich timezone

pablogarciamiranda:


.github/dependabot.yaml
```diff
@@ -1,19 +1,27 @@
----
 version: 2
 updates:
   - package-ecosystem: github-actions
     directory: /
     schedule:
       interval: weekly
+      day: monday
+      time: "08:00"
```
jstudler, line 8:
```suggestion
      time: "04:00"
```

PR #346, opened 2025-06-26
Distinguish if error occurd or if no matching prefix was found

bruelea:
This PR fixes a panic that occurred if now matching parent prefix was found for a prefixclaim.


gdask:
It looks fine.
A regression test was done:

`apiVersion: netbox.dev/v1
kind: PrefixClaim
metadata:
  labels:
    app.kubernetes.io/name: netbox-operator
    app.kubernetes.io/managed-by: kustomize
  name: prefixclaim-parentprefixselector-sample
spec:
  tenant: "MY_TENANT"
  site: "DM-Akron"
  description: "some description"
  comments: "your comments"
  preserveInNetbox: true
  prefixLength: "/31"
  parentPrefixSelector:
    tenant: "MY_TENANT"
    family: "IPv4"
    environment: "Home"
    poolName: "Pool 1"`

After the fix an error "no parent prefix found matching the parenPrefixSelector" is thrown instead of a panic.

PR #395, opened 2025-07-28
Adding automatic CHANGELOG generation

pablogarciamiranda:
Add automatic changelog generation to release workflow

This PR extends the automatic release workflow (#390) to include changelog generation.

### Changes
- Added a step to automatically update CHANGELOG.md when creating releases
- Added script that fetches release notes from GitHub and updates the changelog

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -9,6 +9,3 @@ changelog:
     - title: Version Bumps
       labels:
         - dependencies
-    - title: What's Changed
-      labels:
-        - "*"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,27 +5,35 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [v0.2.5-test] - 2025-07-23
+## [v0.2.8] - 2025-08-04
 
 ## What's Changed
-* Fix iprangeclaim IpRangeName status field yaml/json name by @jstudler in https://github.com/netbox-community/netbox-operator/pull/371
-* Add ownerReference to child resource on update by @jstudler in https://github.com/netbox-community/netbox-operator/pull/369
-* Minor changes to align across controllers by @jstudler in https://github.com/netbox-community/netbox-operator/pull/372
-* Verify prefix length when restoring PrefixClaim from NetBox by @jstudler in https://github.com/netbox-community/netbox-operator/pull/373
-* Change of NetBox Operator IpAddress short name to enhance UX in Kubernetes 1.33+ by @jstudler in https://github.com/netbox-community/netbox-operator/pull/375
-* Add formatting configurations and yamllint action by @jstudler in https://github.com/netbox-community/netbox-operator/pull/376
-* Clean up e2e test files by @jstudler in https://github.com/netbox-community/netbox-operator/pull/382
-
 ### Version Bumps
-* Bump github.com/fxamacker/cbor/v2 from 2.8.0 to 2.9.0 by @dependabot[bot] in https://github.com/netbox-community/netbox-operator/pull/381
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,423 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [v0.2.5-test] - 2025-07-23
```
faebr, line 8: Will this be removed again once the test-release has been deleted? Otherwise we will be missing these diffs in the next release

CHANGELOG.md
```diff
@@ -0,0 +1,423 @@
...
+<!-- Release notes generated using configuration in .github/release.yml at v0.2.5-test -->
+
+## What's Changed
+### Version Bumps
+* Bump github.com/fxamacker/cbor/v2 from 2.8.0 to 2.9.0 by @dependabot[bot] in https://github.com/netbox-community/netbox-operator/pull/381
+* Bump github.com/google/cel-go from 0.25.0 to 0.26.0 by @dependabot[bot] in https://github.com/netbox-community/netbox-operator/pull/380
+* Bump github.com/spf13/pflag from 1.0.6 to 1.0.7 by @dependabot[bot] in https://github.com/netbox-community/netbox-operator/pull/379
+* Bump k8s.io/api from 0.33.2 to 0.33.3 by @dependabot[bot] in https://github.com/netbox-community/netbox-operator/pull/377
+* Bump k8s.io/client-go from 0.33.2 to 0.33.3 by @dependabot[bot] in https://github.com/netbox-community/netbox-operator/pull/378
+### What's Changed
```
faebr, line 21: Why does this release look different than the others? If it will be deleted and not happen again we can leave it but just to understand what the reason is

faebr: ok this is from another mr, added a comment there: https://github.com/netbox-community/netbox-operator/pull/390

scripts/update-changelog-from-release.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Get release data from GitHub
+RELEASE_DATA=$(gh release view "$TAG_NAME" --json tagName,publishedAt,body,url)
+
+# Extract fields
+TAG=$(echo "$RELEASE_DATA" | jq -r '.tagName')
+DATE=$(echo "$RELEASE_DATA" | jq -r '.publishedAt' | cut -d'T' -f1)
+BODY=$(echo "$RELEASE_DATA" | jq -r '.body')
+URL=$(echo "$RELEASE_DATA" | jq -r '.url')
+
```
faebr, line 24: What do you think about adding a check here to see if the fields did not change and we have data?

```suggestion
if [ -z "$TAG" ] || [ -z "$DATE" ] || [ -z "$URL" ]; then
    echo "Release data incomplete. Please check the release format. Exiting."
    exit 1
fi
```

faebr (changes requested):

CHANGELOG.md
pablogarciamiranda: Right now the automation only add new entries. I could look into removing the CHANGELOG.md section whenever a release it's deleted. What do you think? Is it too much?

In any case, if we delete the test release and create a new one, GH will automatically include all those PRs in the next version release.

faebr: I think deleting a release should never happen ( except now when you are testing them and creating test-releases ). So I don't think we need this logic.

.github/release.yml
```diff
@@ -9,6 +9,3 @@ changelog:
     - title: Version Bumps
       labels:
         - dependencies
-    - title: What's Changed
```
faebr, line 12: So why do we remove this in the release notes? Does this mean there will be no more whats changed block anymore?

pablogarciamiranda: This was for the h3 block

faebr (approved):

PR #415, opened 2025-08-17
Add tag support for IPAM/Prefix resources

vaishutin:
This PR introduces support for managing tags on IPAM/Prefix resources in the Netbox Operator.
	•	Extended the CRD to include a tags field in the PrefixSpec.
	•	Implemented tag handling logic in the Netbox client and controller.
	•	Ensures that tags are created/updated/removed accordingly when syncing Prefix objects with NetBox.

Why:
Tags are widely used in NetBox to group and organize objects. Adding tag support to the operator makes Prefix management more consistent with how NetBox is typically used, and allows users to fully automate tagging via Kubernetes manifests.

Notes:
	•	Backwards-compatible change (resources without tags continue to work as before).
	•	Includes CRD schema updates and regeneration of api/v1.


jstudler (changes requested):
Hey @vaishutin

Thanks a lot for your proposal and this MR. I think we can generally allow tags to be added to NetBox Operator resources since it will add value to some users and as long as it will be non-breaking.

After an initial review of this MR I have a few comments and questions:

- Generally, I think it would make sense that this aligns with other usages of NetBox slug/name combinations in the Operator. With site and tenant we use the "name" value and not the "slug" in CRs. Example: https://github.com/netbox-community/netbox-operator/blob/main/api/v1/prefix_types.go#L34-L43
- If we stick to supporting both slug and name, we should verify on the kube-api that the tags either use a slug or a name. This can be done using XValidation rules. Example: `// +kubebuilder:validation:XValidation:rule="(!has(self.parentPrefix) && has(self.parentPrefixSelector)) || (has(self.parentPrefix) && !has(self.parentPrefixSelector))"` https://github.com/netbox-community/netbox-operator/blob/main/api/v1/prefixclaim_types.go#L26
- If we support tags in the Prefix CRs, we should add it as well in the PrefixClaim.
- NetBox IP Ranges and IP Addresses also support tags, I propose to add it as well in IpAddress, IpAddressClaim, IpRange, IpRangeClaim.
- Please add new e2e or integration tests that test the creation of resources with tags.
- If tags are mutable in the Kubernetes CR, we should also test mutating the tags in the e2e or integration tests. If tags are immutable, please add a validation rule. Example: `//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'parentPrefixSelector' is immutable"` https://github.com/netbox-community/netbox-operator/blob/main/api/v1/prefixclaim_types.go#L40 

What do you think?

jstudler:
@vaishutin could you resolve the conflicts of your branch? As a next step, we can then run the CI jobs.

jstudler:
@vaishutin please reach out if there are any questions or we can support here. We'll be happy to review it when the pipeline runs successfully.

PR #436, opened 2025-09-23
start release pipeline on workflow_dispatch

bruelea:
So far the image version and change log were only created after the release was published, so the tag in the release note didn't point to the commit in which the image version is bumped. 

With the new manually triggered workflow the new version and a release summary can be used as parameters to create a new release where the tag points to the commit in which the image is bumped.
This PR is based on the proposal of @alexandernorth and @pablogarciamiranda.

Tested in this fork: https://github.com/bruelea/netbox-operator

Fixed netbox deployment in e2e tests by using bitnamilegacy image repository

### DIFF
--- a/kind/deploy-netbox.sh
+++ b/kind/deploy-netbox.sh
@@ -8,7 +8,7 @@ set -e -o pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 
 # Allow override via environment variable, otherwise fallback to default
-NETBOX_HELM_CHART="${NETBOX_HELM_REPO:-https://github.com}/netbox-community/netbox-chart/releases/download/netbox-7.0.2/netbox-7.0.2.tgz"
+NETBOX_HELM_CHART="${NETBOX_HELM_REPO:-https://github.com}/netbox-community/netbox-chart/releases/download/netbox-5.0.0-beta.169/netbox-5.0.0-beta.169.tgz"
 
 if [[ $# -lt 3 || $# -gt 4 ]]; then
     echo "Usage: $0 <CLUSTER> <VERSION> <NAMESPACE> [--vcluster]"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -25,50 +32,57 @@ jobs:
...
-          if [ -z "$TAG_MESSAGE" ]; then
-            TAG_MESSAGE="Release $TAG_NAME"
+          # Fail if tag already exists
+          if [ -n "$(git tag --list "${{ inputs.version }}")" ]; then
+            echo "Tag "${{ inputs.version }}" already exists. Exit."
+            exit 1
+          else
+            echo "Tag "${{ inputs.version }}" does not exist. Continue."
           fi
+          ./scripts/bump-version.sh "${{ inputs.version }}"
```
alexandernorth, line 44: Why do we not use kustomize here to set the image? Then we don’t need this script

.github/workflows/release.yaml
```diff
@@ -25,50 +32,57 @@ jobs:
...
+          # Read the generated release notes
+          RELEASE_NOTES=$(cat /tmp/release_notes.md)
+
+          # Update changelog with the generated notes
+          echo "" >> CHANGELOG.md
+          echo "## ${{ inputs.version }}" >> CHANGELOG.md
+          echo "" >> CHANGELOG.md
+          echo "$RELEASE_NOTES" >> CHANGELOG.md
+
+          # Also run your existing changelog script if needed
```
alexandernorth, line 84: This comment is probably unneeded? Also do we want to run this script? If we already generated the changelog to the same file the script then uses?

.github/workflows/release.yaml
```diff
@@ -81,23 +95,34 @@ jobs:
...
             --head "$BRANCH_NAME" \
             --base main)
 
           # Extract PR number from URL
           PR_NUMBER=$(echo "$PR_URL" | sed 's/.*\/pull\///')
           echo "Created PR #$PR_NUMBER: $PR_URL"
 
-          # Enable auto-merge and merge the PR
-          gh pr merge "$PR_NUMBER" --auto --squash --delete-branch
+          # Enable merge the PR
```
alexandernorth, line 116:
```suggestion
          # Merge the PR
```

alexandernorth: Also do we not need to enable auto merge?

.github/workflows/release.yaml
```diff
@@ -81,23 +95,34 @@ jobs:
...
+      - name: Create GitHub Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Create release with auto-generated notes
+          gh release create "${{ inputs.version }}" \
+            --title "${{ inputs.version }}" \
+            --notes "${{ inputs.tagMessage }}" \
+            --generate-notes \
+            $PRERELEASE
```
alexandernorth, line 128: Does this variable still exist? It looked like it was removed in the earlier step

scripts/update-changelog-from-release.sh
```diff
@@ -13,14 +13,14 @@ fi
 
 echo "Updating CHANGELOG.md for release $TAG_NAME..."
 
-# Get release data from GitHub
-RELEASE_DATA=$(gh release view "$TAG_NAME" --json tagName,publishedAt,body,url)
+# Get release data from file
+RELEASE_NOTES=$(cat /tmp/release_notes.md)
```
alexandernorth, line 17: Does this file exist?

alexandernorth (commented):
Was doing this from my phone apologies if I didn’t see some reference

bruelea (commented):
Thanks for the suggestions. I cleaned up some parts of the script based on the comments and tested it in the forked repo.

.github/workflows/release.yaml
```diff
@@ -25,50 +32,57 @@ jobs:
...
+
+          # Generate release notes using GitHub API without creating a release
+          if [ -n "$PREVIOUS_TAG" ]; then
+            RELEASE_NOTES=$(gh api repos/:owner/:repo/releases/generate-notes \
+              -f tag_name="${{ inputs.version }}" \
+              -f target_commitish="$(git rev-parse HEAD)" \
+              -f previous_tag_name="$PREVIOUS_TAG" \
+              --jq '.body')
+          else
+            # If no previous tag, generate notes from all commits
```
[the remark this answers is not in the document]
faebr: I was just thinking that if there is an issue in the future with fetching the tags it will recreate everything and that then the issue might go unnoticed, but I guess that one should look at the release when doing it. Could also be left as a comment. But I don't have a strong opinion on removing it so both fine with me

faebr (approved):

PR #437, opened 2025-09-24
Adding app label to spec.selector

pablogarciamiranda:
Simple fix to add the `app.kubernetes.io/name: netbox-operator`  to the spec.selector so it can be used for filtering logs properly.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -16,6 +16,7 @@ spec:
   selector:
     matchLabels:
       control-plane: controller-manager
+      app.kubernetes.io/name: thor-operator
   replicas: 1
   template:
     metadata:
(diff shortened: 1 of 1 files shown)


config/manager/manager.yaml
faebr, line 1: don't you need to add the new label also to the template?

faebr (changes requested):

config/manager/manager.yaml
faebr: see error: https://github.com/netbox-community/netbox-operator/actions/runs/17971459446/job/51114858443?pr=437

faebr (approved):
LGTM
We will need to add a note to the next release notes since this label is not mutable. Which means that if netbox-operator is deployed using for example flux, one needs to manually delete the deployment.

PR #497, opened 2026-01-19
Add support for VLANs

DavideRutigliano:


bruelea:
Hi @DavideRutigliano

Thanks for opening this PR to add support for the vlan model to the NetBox Operator.

I'd suggest to add some e2e tests and to edit the samples so that they can directly be used with the test data of the local setup created by the 'create-kind' command of the make file.

For the vlan model what's the difference between the vlan CR and the vlanclaim CR? Is it the vlanId which will be claimed dynamically by the vlanclaim controller?

jstudler:
@DavideRutigliano could you resolve the conflicts of your branch? As a next step, we can then run the CI jobs.

jstudler:
@DavideRutigliano please reach out if there are any questions or we can support here. We'll be happy to review it when the pipeline runs successfully.

PR #565, opened 2026-05-13
handle iprange overlap error

bruelea:
Due to race conditions ip ranges can get stuck in case they overlap with other ip ranges. This MR will cause the deletion of an iprange cr in case there is an overlap so that the claim controller can claim a new free iprange.

Manual Testing:
Set a break point in the iprange cotroller and manually create an overlapping iprange in NetBox.

Additional Changes:
Change the by default deployed NetBox version for `make create-kind` to 4.4.9.


internal/controller/iprange_controller.go
```diff
@@ -160,12 +160,15 @@ func (r *IpRangeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 		return ctrl.Result{}, err
 	}
 
+	overlapErr := &api.OverlapError{}
```
alexandernorth: why not have this in the if-block on line 166?

pkg/netbox/api/ip_range.go
```diff
@@ -31,6 +32,35 @@ import (
...
+type OverlapError struct {
+	Range string
+	VRF   string
+	Msg   string
+}
+
+var overlapRegex = regexp.MustCompile(`(?i)Defined addresses overlap with range (.*?) in VRF (.*)$`)
+
+func (e *OverlapError) Error() string {
+	return e.Msg
```
alexandernorth, line 42: maybe this should take advantage of the `Range`/`VRF` fields? Or we could drop them if they bring no extra value

internal/controller/iprange_controller.go
bruelea: moved it into the if-block

pkg/netbox/api/ip_range.go
bruelea: Since the fields are not used here, I dropped them to simplify the error type.

alexandernorth (approved):
lgtm

PR #571, opened 2026-05-27
Add scheduled reconciliation to all controllers

MaIT-HA:
This PR introduces optional scheduled reconciliation across the operator’s controllers, configured via a cron expression plus a jitter window to spread load over time.

Changes:

- Add RECONCILE_SCHEDULE (cron) and RECONCILE_JITTER (duration) to operator config parsing.
- Implement a shared scheduler helper and wire it into each controller’s reconcile defer-path when no other requeue is requested.
- Add kind overlay patch to set schedule/jitter env vars (and update kind kustomization).

### DIFF
--- a/internal/controller/ipaddress_controller.go
+++ b/internal/controller/ipaddress_controller.go
@@ -30,6 +30,8 @@ import (
 	"github.com/netbox-community/netbox-operator/pkg/config"
 	"github.com/netbox-community/netbox-operator/pkg/netbox/api"
 	"github.com/netbox-community/netbox-operator/pkg/netbox/models"
+	"github.com/netbox-community/netbox-operator/pkg/scheduler"
+
 	"github.com/swisscom/leaselocker"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -79,14 +81,18 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// merge-patch diff captures every change (IpAddressId, conditions, etc.).
 	statusBase := o.DeepCopy()
 
-	// cancelLock stops the lease renewal goroutine on early returns (lease expires naturally).
-	// Explicit cancelLock()+UnlockWithRetry() runs inline after the critical section.
-	var cancelLock context.CancelFunc
-
 	// Defer status update to ensure it happens regardless of how we exit
 	defer func() {
 		reconcileResult, reconcileErr = r.updateStatus(ctx, o, statusBase, reconcileResult, reconcileErr)
+		if reconcileErr == nil && reconcileResult.IsZero() {
+			reconcileResult, reconcileErr = scheduler.CalculateNextReconcile(ctx)
+		}
+		logger.Info("reconcile loop finished")
 	}()
+
+	// cancelLock stops the lease renewal goroutine on early returns (lease expires naturally).
+	// Explicit cancelLock()+UnlockWithRetry() runs inline after the critical section.
+	var cancelLock context.CancelFunc
 
 	// if being deleted
 	if !o.DeletionTimestamp.IsZero() {
--- a/internal/controller/ipaddress_controller.go
+++ b/internal/controller/ipaddress_controller.go
@@ -84,8 +84,19 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// Defer status update to ensure it happens regardless of how we exit
 	defer func() {
 		reconcileResult, reconcileErr = r.updateStatus(ctx, o, statusBase, reconcileResult, reconcileErr)
+		logger.Info("evaluating scheduled reconciliation",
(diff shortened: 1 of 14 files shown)


pkg/config/config.go
```diff
@@ -103,3 +133,43 @@ func GetProtocol() string {
...
+			return fmt.Errorf("invalid reconcile jitter %q: %w", c.ReconcileJitterRaw, err)
+		}
+
+		if c.ReconcileJitterDuration < 0 {
+			return fmt.Errorf("invalid reconcile jitter %q: must be greater than or equal to 0", c.ReconcileJitterRaw)
+		}
+	}
+
+	// Parse cron schedule
+	c.ReconcileSchedule = nil // default
```
bruelea:
```suggestion
```

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2024 Swisscom (Schweiz) AG.
```
bruelea:
```suggestion
Copyright 2026 Swisscom (Schweiz) AG.
```

pkg/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,101 @@
+/*
+Copyright 2024 Swisscom (Schweiz) AG.
```
bruelea:
```suggestion
Copyright 2026 Swisscom (Schweiz) AG.
```

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,69 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+func CalculateNextReconcile(ctx context.Context) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	// do not reschedule if no schedule is defined
+	if config.GetOperatorConfig().ReconcileSchedule == nil {
+		logger.Info("Scheduled reconciliation disabled: no reconcile schedule configured")
```
bruelea: If this log is needed I'd prefer to move it to config.go, to make sure it's not logged every time the reconcile function is called. For all logs related to scheduled reconciliation the debug level could be used.

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,69 @@
...
+	// Calculate duration till next reconciliation and add jitter
+	jitter := getJitterDuration()
+	now := time.Now()
+	nextRunWithJitter := config.GetOperatorConfig().ReconcileSchedule.Next(now).Sub(now) + jitter
+	if nextRunWithJitter < 0 {
+		nextRunWithJitter = 0
+	}
+
+	logger.Info("Calculated next reconciliation delay",
+		"nextRunWithJitter", nextRunWithJitter.String())
```
bruelea:
```suggestion
```

internal/controller/ipaddress_controller.go
```diff
@@ -79,15 +81,30 @@ func (r *IpAddressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	statusBase := o.DeepCopy()
 
-	// cancelLock stops the lease renewal goroutine on early returns (lease expires naturally).
-	// Explicit cancelLock()+UnlockWithRetry() runs inline after the critical section.
-	var cancelLock context.CancelFunc
-
 	// Defer status update to ensure it happens regardless of how we exit
 	defer func() {
 		reconcileResult, reconcileErr = r.updateStatus(ctx, o, statusBase, reconcileResult, reconcileErr)
+		logger.Info("evaluating scheduled reconciliation",
```
bruelea: Because the time until the next reconciliation is already logged in the scheduler, the logs here can be removed.

bruelea (changes requested):
Looks good! In some the operators at the end of the reconcile function there's a line `logger.Info("reconcile loop finished")`. This can now be removed, because it's now part of the defer statement which calculates the delay to the next reconciliation.

pkg/config/config.go
MaIT-HA: coud you tell me why you think we don't need this?

bruelea: If c.ReconcileSchedule is not set it will be nil automatically.

MaIT-HA: okay, I thought the guard would make sense just in case the if the function was called again.

pkg/scheduler/scheduler.go
```diff
@@ -46,11 +45,8 @@ func CalculateNextReconcile(ctx context.Context) (ctrl.Result, error) {
 		nextRunWithJitter = 0
 	}
 
-	logger.Info("Calculated next reconciliation delay",
-		"nextRunWithJitter", nextRunWithJitter.String())
-
 	if nextRunWithJitter > 0 {
-		logger.Info("Scheduled next reconciliation",
+		logger.V(1).Info("Scheduled next reconciliation",
```
bruelea: For debug level we usually use V(4).

bruelea (approved):

PR #572, opened 2026-05-29
Fix wrongfully assigned parentPrefix that is already exhausted.

jstudler:
Add a test that reproduces a bug we've seen in production:

1. Creates 4.0.2.0/24 parent prefix → asserts Ready
2. Creates two /25 claims → both assigned from 4.0.2.0/24
3. Creates third /25 claim → gets exhaustion warning event
4. Creates 4.0.3.0/24 parent prefix → asserts Ready
5. Third claim doesn't become Ready, instead it will set the following condition:
```
  - lastTransitionTime: "2026-05-29T06:10:23Z"
    message: 'The parent prefix was selected successfully, parentPrefix is selected:
      4.0.2.0/24'
    observedGeneration: 1
    reason: ParentPrefixSelected
    status: "True"
    type: ParentPrefixSelected
```

Chainsaw test for the expected behaviour is added to this PR. When the code is fixed, the test should succeed.

### DIFF
--- a/kind/load-local-data-job/main.py
+++ b/kind/load-local-data-job/main.py
@@ -225,6 +225,7 @@ class Prefix:
 
 scopeId = nb.dcim.sites.get(name="MY_SITE").id
 
+# Note: 4.0.0.0/8 (4.x.y.z) prefixes are used by e2e tests that dynamically create the test data using Prefix custom resources.
 prefixes = [
     Prefix(
         prefix="2.0.0.0/16",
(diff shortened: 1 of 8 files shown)


jstudler:
Fixed the error handling in GetAvailablePrefixesByParentPrefixSelector. Before this fix, all prefixes matching the parentPrefixSelector were returned as valid. Now, only the prefixes matching the parentPrefixSelector that still have sufficient capacity will be returned. Since in the PrefixClaim Controller it just selected the first in the list (https://github.com/netbox-community/netbox-operator/blob/main/internal/controller/prefixclaim_controller.go#L157-L159 ) the selection was dependent on the order of the CIDRs in NetBox. So if a new PrefixClaim was added to NetBox with a lower CIDR than the exhausted, it would still work (in the example of the PR description that'd mean in Step 4, use 4.0.1.0/24 instead).

bruelea (approved):
Looks good!

PR #606, opened 2026-07-31
Add AsnClaim and Asn

jstudler:
## Add ASN and AsnClaim support

Adds ASN (Autonomous System Number) management to the NetBox Operator, following the same claim/resource pattern already used for IP addresses, prefixes and IP ranges.

### New CRDs

- `Asn` — represents a single ASN in NetBox.
- `AsnClaim` — claims an available ASN from a NetBox ASN Range via `parentAsnRange` (matched by range name) and creates an owned `Asn` CR. Status exposes the assigned `asn` and `asnName`.

### Controllers

- `AsnReconciler` — reserves/updates the ASN in NetBox, finalizer-based cleanup (`asn.netbox.dev/finalizer`), lease-locks the parent ASN range while the ASN is not yet Ready, tracks managed custom fields via annotation, reports conditions and events.
- `AsnClaimReconciler` — allocates an ASN from the parent range, restores a previously assigned ASN by hash before falling back to a new allocation, resolves the effective RIR, syncs mutable fields down to the owned `Asn`, lease-locks per range to avoid races between concurrent claims.

### RIR assignment

NetBox requires every ASN to belong to a RIR. Rather than deriving it implicitly, the RIR is exposed as a spec field:

- `Asn.spec.rir` is **required** and resolved by name against NetBox.
- `AsnClaim.spec.rir` is **optional**. When omitted, the RIR of the parent ASN Range is inherited and written to the generated `Asn` CR.

Both fields are mutable and neither is part of the restoration hash, so changing the RIR never causes an ASN to be re-claimed. The `AsnClaim` controller re-resolves the effective RIR on every reconcile and propagates it to the `Asn`, which then updates the ASN in NetBox.

The RIR is set on both the create and the update path, since NetBox treats an ASN update as a full replacement and would otherwise clear it.

### NetBox client

New methods on `NetboxCompositeClient`: `ReserveOrUpdateAsn`, `DeleteAsn`, `RestoreExistingAsnByHash`, `GetAvailableAsnByClaim`, `GetAsnRangeRirName`, plus RIR-by-name and ASN-range-by-name lookups. New `ASN` / `ASNClaim` / `Rir` models, `IpamRirsList` added to the `IpamAPI` interface, and regenerated mocks for the `IpamAsns*` / `IpamAsnRanges*` / `IpamRirs*` API surface.

### Tests & docs

- Unit tests for controllers, helpers and the NetBox client, including a regression test asserting that `spec.rir` does not influence the restoration hash.
- Chainsaw e2e tests under `tests/e2e/asn`: apply/update, restore, range-exhausted, owner-reference and RIR-override (inherit, then override) cases; local test data (two RIRs + ASN ranges) seeded in the kind data-load job.
- README section "ASN Management" documenting the `Asn`/`AsnClaim` model, `parentAsnRange` semantics, RIR assignment and restoration.

### Behaviour notes

- Restoration hash makes ASN assignment sticky across CR re-creation; a hash mismatch on a not-yet-reserved `Asn` deletes the CR.
- `preserveInNetbox: true` keeps the ASN in NetBox after CR deletion, enabling later reclaim.
- `asn` and `tenant` immutability is enforced by CEL validation rules; `rir` is deliberately mutable.


internal/controller/asn_controller.go
```diff
@@ -0,0 +1,276 @@
...
+//+kubebuilder:rbac:groups=netbox.dev,resources=asns/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=netbox.dev,resources=asns/finalizers,verbs=update
+//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *AsnReconciler) Reconcile(ctx context.Context, req ctrl.Request) (reconcileResult ctrl.Result, reconcileErr error) {
+	logger := log.FromContext(ctx)
+
+	logger.Info("reconcile loop started")
```
alexandernorth, line 66: This should probably be at a different logging level, e.g. Debug

internal/controller/asn_controller.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+
+		if err = r.Update(ctx, o); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		return ctrl.Result{}, nil
+	}
+
+	// if PreserveInNetbox flag is false then register finalizer if not yet registered
```
alexandernorth, line 112: should we also remove the finaliser if the flag is set? This case is already handled in the logic, but it's something we could consider (e.g. flag toggled after create)

internal/controller/asn_controller.go
```diff
@@ -0,0 +1,276 @@
...
+	asnModel, err := generateNetboxAsnModelFromAsnSpec(&o.Spec, req, annotations[AsnManagedCustomFieldsAnnotationName])
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	netboxAsnModel, statusUpToDate, err := r.NetboxClient.ReserveOrUpdateAsn(ctx, asnModel, o)
+	if err != nil {
+		if errors.Is(err, api.ErrRestorationHashMismatch) && o.Status.AsnId == 0 {
+			logger.Info("restoration hash mismatch, deleting ASN custom resource", "asn", o.Spec.Asn)
+			if deleteErr := r.Delete(ctx, o); deleteErr != nil {
```
alexandernorth, line 137: Do we want to do this only if it is owned by a claim?

internal/controller/asn_controller.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	// 2.1 update annotations
+	// the accessor returns the annotation map of o itself, so the merge-patch base has to be
+	// snapshotted before that map is mutated, otherwise the patch is empty and never applied
+	patch := client.MergeFrom(o.DeepCopy())
+
+	if annotations == nil {
+		annotations = make(map[string]string, 1)
+	}
```
alexandernorth, line 158: I wonder if it is not clearer to put this logic where `annotations` is declared/set for the first time

internal/controller/asn_controller.go
```diff
@@ -0,0 +1,276 @@
...
+	if err := r.Patch(ctx, o, patch); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// 3. update status fields
+	o.Status.AsnId = int64(netboxAsnModel.Id)
+	o.Status.AsnUrl = config.GetBaseUrl() + "/ipam/asns/" + strconv.FormatInt(int64(netboxAsnModel.Id), 10)
+	if netboxAsnModel.LastUpdated.Get() != nil {
+		o.Status.LastUpdated = metav1.NewTime(*netboxAsnModel.LastUpdated.Get())
+	}
```
alexandernorth, line 178: Do we want to put this earlier (after the actual API call) else errors which occur between these two points will result in status not being updated

internal/controller/asnclaim_controller.go
```diff
@@ -0,0 +1,265 @@
...
+//+kubebuilder:rbac:groups=netbox.dev,resources=asnclaims/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=netbox.dev,resources=asnclaims/finalizers,verbs=update
+//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *AsnClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (reconcileResult ctrl.Result, reconcileErr error) {
+	logger := log.FromContext(ctx)
+
+	logger.Info("reconcile loop started")
```
alexandernorth, line 60: as previous, should be debug level

internal/controller/asnclaim_helpers.go
```diff
@@ -0,0 +1,82 @@
...
+	return fmt.Sprintf("%x", sha1.Sum([]byte(rd.Namespace+rd.Name+rd.ParentAsnRange+rd.Tenant)))
+}
+
+type AsnClaimRestorationData struct {
+	// only use immutable fields
+	Namespace      string
+	Name           string
+	ParentAsnRange string
+	Tenant         string
+}
```
alexandernorth, line 82: I would add a function to AsnClaimRestorationData which produces the hash so that this logic is kept together

internal/controller/asnclaim_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	err = reconcileErr
+
+	// Ensure status update is always called, even on early returns
+	defer func() {
+		if apierrors.IsConflict(err) {
+			result, err = IgnoreDomainError(result, err)
+			return
+		}
+		statusBase.SetResourceVersion(claim.GetResourceVersion())
+		statusPatch := client.MergeFrom(statusBase)
```
alexandernorth, line 218: Copying the live resourceVersion onto the base drops it from the patch body, so
the apiserver can't reject a stale status write and we just overwrite. Can we
keep that check instead?

```suggestion
		statusPatch := client.MergeFromWithOptions(statusBase, client.MergeFromWithOptimisticLock{})
```

pkg/netbox/api/asn.go
```diff
@@ -0,0 +1,240 @@
...
+	// else and must never be adopted.
+	restorationHashKey := config.GetOperatorConfig().NetboxRestorationHashFieldName
+	if asn.Metadata != nil {
+		if restorationHash, ok := asn.Metadata.Custom[restorationHashKey]; ok {
+			cfHash, cfOk := asnToUpdate.CustomFields[restorationHashKey]
+			if !cfOk || cfHash == nil || cfHash == "" || cfHash != restorationHash {
+				return nil, false, fmt.Errorf("%w, assigned ASN %d", ErrRestorationHashMismatch, asn.Asn)
+			}
+		}
+	}
```
alexandernorth, line 115: I’m not sure this matches the comment. If Metadata is nil or the hash is missing we skip the check and adopt it. Can we make those cases explicit?


```suggestion
  // If the desired ASN carries a restoration hash, the ASN in NetBox must carry exactly
  // the same one. An ASN without a hash, or with a different hash, belongs to somebody
  // else and must never be adopted.
  // A nil metadata, missing key, or empty value means the desired ASN does not carry a
  // hash; that is not a restore, so the existing object may be updated.
  restorationHashKey := config.GetOperatorConfig().NetboxRestorationHashFieldName
  desiredHash := ""
  if asn.Metadata != nil {
  	desiredHash = asn.Metadata.Custom[restorationHashKey]
  }
  if desiredHash != "" {
  	netboxHash, _ := asnToUpdate.CustomFields[restorationHashKey].(string)
  	if netboxHash != desiredHash {
  		return nil, false, fmt.Errorf("%w, assigned ASN %d", ErrRestorationHashMismatch, asn.Asn)
  	}
  }
  ```

File: pkg/netbox/api/rir.go [hunk 10]
```diff
@@ -0,0 +1,49 @@
...
+	closeFunc, handleErr := handleHTTPResponse(httpResp, execErr, http.StatusOK, "fetch RIR details")
+	if closeFunc != nil {
+		defer func() { err = errors.Join(err, closeFunc()) }()
+	}
+	if handleErr != nil {
+		return nil, handleErr
+	}
+
+	if len(result.Results) == 0 {
+		return nil, utils.NetboxNotFoundError("RIR '" + name + "'")
```
alexandernorth, line 41:
```suggestion
		return nil, utils.NetboxNotFoundError(fmt.Sprintf("RIR '%s'", name))
```

internal/controller/asn_controller.go
jstudler: Agree, but it's aligned with the other controllers, I'd prioritize consistency here. What do you think?

internal/controller/asnclaim_controller.go
jstudler: let's have this discussion in the other thread.

internal/controller/asn_controller.go
jstudler: I kept it aligned with the other non-Claim Controllers. If we want to change this, I'd change it in all controllers simultaneously.

internal/controller/asn_controller.go
jstudler: Agree, that would be a good and important failsafe. It actually goes against good practices to change the spec of the watched resource (by deleting it). Propose to create a separate PR to introduce the ownerReference kind check for all controllers.

internal/controller/asn_controller.go
jstudler: Yes, could be better. Same as with the others, it aligns with the other controllers. Separate PR?

internal/controller/asn_controller.go
jstudler: So updating the status before annotations? Could make sense because if we lose the status we might lose the reference to the resource in NetBox. What do you think @bruelea ?

internal/controller/asnclaim_helpers.go
jstudler: This also aligns with other controllers.

jstudler:
Thanks @alexandernorth for your feedback. Since all the open comments are things that apply to all controllers: Would you agree in merging this one as is and adding a new PR for the refactorings in all controllers? I'd prefer to have the different Controllers aligned instead of fixing it only in here.

PR #607, opened 2026-07-31
Add support for L2VPN

flynn-nrg:
## Add L2VPN and L2VPNClaim support

Adds L2VPN (Layer 2 VPN) management to the NetBox Operator, following the same claim/resource pattern already used for IP addresses, prefixes, IP ranges and ASNs.

### New CRDs

- `L2VPN` — represents a single L2VPN in NetBox (name, type, identifier/VNI, tenant, comments, description, custom fields, preserveInNetbox).
- `L2VPNClaim` — claims an L2VPN identifier (VXLAN VNI) either as an exact `identifier` or from an `identifierRangeStart`/`identifierRangeEnd` range, and creates an owned `L2VPN` CR. `identifier` and the range fields are mutually exclusive (CEL validation) and immutable once set.

### Controllers

- `L2VPNReconciler` — reserves/updates the L2VPN in NetBox (`ReserveOrUpdateL2VPN`), finalizer-based cleanup (`l2vpn.netbox.dev/finalizer`), lease-locks the parent identifier range while the L2VPN is not yet Ready, reports conditions and events.
- `L2VPNClaimReconciler` — restores a previously assigned identifier by hash before falling back to the explicit `identifier` or a new range-based allocation, syncs mutable fields down to the owned `L2VPN`, lease-locks per range to avoid races between concurrent claims (`l2vpnclaim.netbox.dev/finalizer`).

### NetBox client

New methods on `NetboxCompositeClient`: `ReserveOrUpdateL2VPN`, `DeleteL2VPN`, `RestoreExistingL2VPNByHash`, `GetAvailableL2VPNIdentifierByClaim` (scans existing L2VPNs, since NetBox has no dedicated "available identifiers" endpoint for L2VPN). New `L2VPN`/`L2VPNClaim` models and regenerated mocks for the `VpnL2vpns*` API surface. The `netboxOperatorRestorationHash` custom field is extended to cover `vpn.l2vpn` in the kind data-load job.

### Tests & docs

- Unit tests for the NetBox client (`pkg/netbox/api/l2vpn_test.go`, `l2vpn_claim_test.go`) and for the controllers (`internal/controller/l2vpn_controller_test.go`, `l2vpnclaim_controller_test.go`), covering reservation, update, restoration-hash mismatch, reserve failure, range allocation, restore-by-hash, and lease-lock contention.
- Chainsaw e2e tests under `tests/e2e/l2vpn`: explicit/range apply-update, range restore, range-exhausted, invalid tenant/custom-field, and owner-reference cases.
- Sample manifests under `config/samples/`.
- README section "L2VPN Management" documenting the `L2VPN`/`L2VPNClaim` model and usage.

### Behaviour notes

- Restoration hash makes identifier assignment sticky across CR re-creation; `preserveInNetbox: true` keeps the L2VPN in NetBox after CR deletion, enabling later reclaim.
- `name`, `type`, `identifier` and `tenant` immutability enforced via CEL validation rules on both `L2VPN` and `L2VPNClaim`.

---

This was done with the help of Claude. I haven't seen anything in the docs about it so I assume it's fine.


bruelea:
Hi @flynn-nrg, thanks for opening a PR. 

If a l2vpnclaim CR is created with the "identifier" field, is it equivalent to creating a l2vpn CR?
We'll follow up with a more detailed review.

flynn-nrg:
> Hi @flynn-nrg, thanks for opening a PR.
> 
> If a l2vpnclaim CR is created with the "identifier" field, is it equivalent to creating a l2vpn CR? We'll follow up with a more detailed review.

It follows the same pattern by the other claims like IpAddressClaim, so it will compute a restoration hash first and check if the matching object already exists. If not and there is a identifier it will be used directly to create a child L2VPN CR.

api/v1/l2vpn_types.go
```diff
@@ -0,0 +1,167 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// L2VPNSpec defines the desired state of L2VPN
+type L2VPNSpec struct {
+	// The name of the L2VPN in NetBox
+	// Field is immutable, required
+	//+kubebuilder:validation:Required
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'name' is immutable"
+	Name string `json:"name"`
```
bruelea: could metadata.Name be used instead?

pkg/netbox/api/l2vpn.go
```diff
@@ -0,0 +1,207 @@
...
+	resp, err = c.updateL2VPN(ctx, l2vpnId, desiredL2VPN)
+	if err != nil {
+		return nil, false, err
+	}
+	return resp, false, nil
+}
+
+func (c *NetboxCompositeClient) getL2VPN(ctx context.Context, l2vpn *models.L2VPN) (*v4client.PaginatedL2VPNList, error) {
+	req := c.clientV4.VpnAPI.VpnL2vpnsList(ctx).
+		Name([]string{l2vpn.Name})
```
bruelea: If the l2vpn are filtered by name, different CR's can point to the same l2vpn in NetBox. What are the unique values the l2vpns could be filterd by?

internal/controller/l2vpnclaim_controller.go
```diff
@@ -0,0 +1,326 @@
...
+		r.EventStatusRecorder.Report(ctx, claim, netboxv1.ConditionL2VPNClaimReadyFalse, corev1.EventTypeWarning, reconcileErr)
+	}
+
+	return result, err
+}
+
+// tryLockOnRange serializes concurrent range-based claims against the same
+// identifier range. Returns a nil locker and nil error when the claim uses an
+// explicit identifier instead of a range, since there's no shared pool to lock.
+func (r *L2VPNClaimReconciler) tryLockOnRange(ctx context.Context, o *netboxv1.L2VPNClaim) (ll *leaselocker.LeaseLocker, cleanup context.CancelFunc, res ctrl.Result, err error) {
```
bruelea: The identifier range can be set be the user, so there could be overlaps. If the identifiers are only unique per l2vpn types the LeaseLockName could be the type. If they need to be unique across type, the LeaseLockName could be "l2vpn" to make sure that different claims cannot get the same identifier assigned. Also for a the claims with explicit identifier there can be a lock, to make sure it does not overlap with a l2vpn which was claimed by a l2vpnclaim with a range.

bruelea (commented):
Currently, different CR's can point to the same l2vpn in NetBox which could be fixed by changing the list option used to get the l2vpn for a CR from NetBox.

api/v1/l2vpn_types.go
flynn-nrg: Yes, done in https://github.com/netbox-community/netbox-operator/pull/607/commits/729792080c4ab23338ad3b8114f9c1b2cc9efd9e

internal/controller/l2vpnclaim_controller.go
flynn-nrg: Reworked the code so that a single lease lock name and avoid this scenario: https://github.com/netbox-community/netbox-operator/pull/607/commits/e0e8c27977abb84684305ad56e91e163cd7d6454

pkg/netbox/api/l2vpn.go
flynn-nrg: Have changed that part to have the adapter expose the identifier, which is now the filter criterion and prevents two CRs with a shared or duplicate name from colliding: https://github.com/netbox-community/netbox-operator/pull/607/commits/13997fbd1899f60b918cc5c92c3fd3ebf07109bd

pkg/netbox/api/l2vpn_claim.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		for i := range resp.Results {
+			if !visit(&resp.Results[i]) {
+				return nil
+			}
+		}
+
+		offset += int32(len(resp.Results))
+		if offset >= resp.Count || len(resp.Results) == 0 {
```
alexandernorth: This check looks to me like it might be incorrect - if there are many pages of results we won't get past the third page as offset will be >= count (limited to 100 results

pkg/netbox/api/l2vpn_claim.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// RestoreExistingL2VPNByHash searches for an existing NetBox L2VPN whose
+// restoration hash custom field matches hash. Returns nil if none is found.
+func (c *NetboxCompositeClient) RestoreExistingL2VPNByHash(ctx context.Context, hash string) (*models.L2VPN, error) {
+	hashKey := config.GetOperatorConfig().NetboxRestorationHashFieldName
+
+	var found *v4client.L2VPN
+	err := c.forEachL2VPN(ctx, func(l2vpn *v4client.L2VPN) bool {
+		if l2vpn.CustomFields == nil {
```
alexandernorth:
```suggestion
		if len(l2vpn.CustomFields) == 0 {
```

internal/controller/l2vpn_controller.go
```diff
@@ -0,0 +1,354 @@
...
+	// if being deleted
+	if !o.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(o, L2VPNFinalizerName) {
+			return ctrl.Result{}, nil
+		}
+
+		if !o.Spec.PreserveInNetbox {
+			if o.Status.L2VPNId > math.MaxInt32 {
+				return ctrl.Result{}, fmt.Errorf("reconciliation of l2vpns with id's larger than 2147483647 is not supported")
+			}
```
alexandernorth: This kind of bounds check would fit better in the `NetboxClient.DeleteL2VPN` function

api/v1/l2vpnclaim_types.go
```diff
@@ -0,0 +1,176 @@
...
+	//+kubebuilder:validation:Enum=vxlan;vxlan-evpn
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'type' is immutable"
+	Type string `json:"type"`
+
+	// The exact VNI to claim. Mutually exclusive with identifierRangeStart/identifierRangeEnd.
+	// Field is immutable, not required, range from 4000-16777215
+	//+kubebuilder:validation:Minimum=4000
+	//+kubebuilder:validation:Maximum=16777215
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'identifier' is immutable"
+	Identifier int64 `json:"identifier,omitempty"`
```
alexandernorth: Could we apply the same logic from https://github.com/netbox-community/netbox-operator/pull/618#discussion_r3864604496 such that we only have a range start/end field?

internal/controller/l2vpn_controller.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+// getLeaseLockerNSNandOwner returns the lease lock identity for the L2VPN's
+// owning claim, against the same shared lock (l2vpnIdentifierLockName) the
+// L2VPNClaim controller used while assigning this L2VPN's identifier, so
+// NetBox reservation stays serialized against concurrent claims for the
+// duration of the create.
+func (r *L2VPNReconciler) getLeaseLockerNSNandOwner(ctx context.Context, o *netboxv1.L2VPN) (nsn types.NamespacedName, owner string, identifierDesc string, err error) {
+	orLookupKey := types.NamespacedName{
+		Name:      o.ObjectMeta.OwnerReferences[0].Name,
```
alexandernorth, line 338: I would add a guard against dereferencing an array which has not been checked within this function already

alexandernorth (changes requested):
Hi @flynn-nrg thank you for your contribution. I had a quick look at the main logic and have the following comments, please reach out if there is something which is not clear

pkg/netbox/api/l2vpn_claim.go
flynn-nrg: This is fixed in https://github.com/netbox-community/netbox-operator/pull/607/commits/08127863829bda2e580752d6bb5649152aacbd8d

pkg/netbox/api/l2vpn_claim.go
flynn-nrg: Applied.

internal/controller/l2vpn_controller.go
flynn-nrg: Moved in https://github.com/netbox-community/netbox-operator/pull/607/commits/08127863829bda2e580752d6bb5649152aacbd8d

api/v1/l2vpnclaim_types.go
flynn-nrg: Removed identifier in https://github.com/netbox-community/netbox-operator/pull/607/commits/08127863829bda2e580752d6bb5649152aacbd8d

internal/controller/l2vpn_controller.go
flynn-nrg: Added a guard in https://github.com/netbox-community/netbox-operator/pull/607/commits/08127863829bda2e580752d6bb5649152aacbd8d

pkg/netbox/api/l2vpn.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return nil, false, err
+	}
+	return resp, false, nil
+}
+
+func (c *NetboxCompositeClient) getL2VPN(ctx context.Context, l2vpn *models.L2VPN) (*v4client.PaginatedL2VPNList, error) {
+	req := c.clientV4.VpnAPI.VpnL2vpnsList(ctx).
+		Identifier([]int32{int32(l2vpn.Identifier)})
+	resp, httpResp, err := req.Execute()
```
bruelea, line 131:
```suggestion
func (c *NetboxCompositeClient) getL2VPN(ctx context.Context, l2vpn *models.L2VPN) (resp *v4client.PaginatedL2VPNList, err error) {
	req := c.clientV4.VpnAPI.VpnL2vpnsList(ctx).
		Identifier([]int32{int32(l2vpn.Identifier)})
	var httpResp *http.Response
	resp, httpResp, err = req.Execute()
```
To make sure errClose in the deferred function is not lost.

api/v1/l2vpn_types.go
```diff
@@ -0,0 +1,162 @@
...
+	// Field is immutable, required
+	//+kubebuilder:validation:Required
+	//+kubebuilder:validation:Enum=vxlan;vxlan-evpn
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="Field 'type' is immutable"
+	Type string `json:"type"`
+
+	// The VNI to be assigned to this L2VPN in NetBox
+	// Field is immutable, required, range from 4000-16777215
+	//+kubebuilder:validation:Required
+	//+kubebuilder:validation:Minimum=4000
```
bruelea, line 36: What's the reason for setting the miminum to 4000?

bruelea (commented):
I've tested the latest changes, works as expected. Only added two minor questions/suggestions.

api/v1/l2vpn_types.go
flynn-nrg: At some places I've worked we used to leave the bottom 4000 ids reserved but I've set it to 0 as the RFC does not impose such a restriction.

https://github.com/netbox-community/netbox-operator/pull/607/commits/b736be408f877781ef6838d49cd6b7318d1ec31a

alexandernorth (approved):
thanks for your work!

bruelea (approved):

PR #618, opened 2026-08-17
Add Vlan and VlanClaim support

flynn-nrg:
## Add Vlan and VlanClaim support

Adds VLAN management to the NetBox Operator, following the same claim/resource pattern already used for IP addresses, prefixes, and IP ranges.

### New CRDs

- `Vlan` represents a single VLAN in NetBox (name, vid, site, tenant, status, comments, description, custom fields, preserveInNetbox).
- `VlanClaim` claims a VLAN ID either as an exact `vid` or from a `vidRangeStart`/`vidRangeEnd` range, and creates an owned `Vlan` CR. `vid` and the range fields are mutually exclusive (CEL validation) and immutable once set. VID allocation is scoped to `.spec.site`, since VLAN IDs are commonly only unique within a site.

Only `tenant` and `site` references are supported for now (both reuse existing lookup helpers).

### Controllers

- `VlanReconciler` reserves/updates the VLAN in NetBox (`ReserveOrUpdateVlan`), finalizer-based cleanup (`vlan.netbox.dev/finalizer`), lease-locks the parent VID range while the Vlan is not yet Ready, reports conditions and events.
- `VlanClaimReconciler` restores a previously assigned VID by hash before falling back to the explicit `vid` or a new range-based allocation, syncs mutable fields down to the owned `Vlan`, lease-locks per site+range to avoid races between concurrent claims (`vlanclaim.netbox.dev/finalizer`).

### NetBox client

New methods on `NetboxCompositeClient`: `ReserveOrUpdateVlan`, `DeleteVlan`, `RestoreExistingVlanByHash`, `GetAvailableVlanByClaim` (scans existing VLANs scoped to the claim's site, since NetBox has no dedicated "available VLAN IDs" endpoint outside of VLAN Groups). VLAN support was added directly to the existing `IpamAPI` interface/adapter. New `Vlan`/`VlanClaim` models and regenerated mocks for the `IpamVlans*` API surface. The `netboxOperatorRestorationHash` custom field is extended to cover `ipam.vlan` in the kind data-load job.

### Tests & docs

- Unit tests for the NetBox client (`pkg/netbox/api/vlan_test.go`, `vlan_claim_test.go`) and for the controllers (`internal/controller/vlan_controller_test.go`, `vlanclaim_controller_test.go`), covering reservation, update, restoration-hash mismatch, reserve failure, range allocation, restore-by-hash, site-scoping, and lease-lock contention.
- Chainsaw e2e tests under `tests/e2e/vlan`: explicit/range apply-update, range restore, range-exhausted, invalid tenant/custom-field, and owner-reference cases.
- Sample manifests under `config/samples/`.
- README section "VLAN Management" documenting the `Vlan`/`VlanClaim` model and usage.

### Behaviour notes

- Restoration hash makes VID assignment sticky across CR re-creation; `preserveInNetbox: true` keeps the Vlan in NetBox after CR deletion, enabling later reclaim.
- `name`, `vid`, `site`, and `tenant` immutability enforced via CEL validation rules on both `Vlan` and `VlanClaim`.
- The VID-range lease-lock name is derived from `site`, which (unlike a controlled enum) is free text, so it's sanitized to satisfy the Lease resource's DNS-1123 naming requirements.

---

This was done with the help of Claude.


pkg/netbox/api/vlan_claim.go
```diff
@@ -0,0 +1,160 @@
...
+		if httpResp != nil && httpResp.Body != nil {
+			_, _ = io.ReadAll(httpResp.Body)
+			closeErr := httpResp.Body.Close()
+			err = errors.Join(err, closeErr)
+		}
+		if httpResp == nil {
+			return fmt.Errorf("failed to list vlans: %w", err)
+		}
+		if httpResp.StatusCode != http.StatusOK {
+			return fmt.Errorf("failed to list vlans: status %d", httpResp.StatusCode)
```
bruelea: Here the response body could be added to the error message, for some information about why the request failed.

config/samples/netbox_v1_vlanclaim_range.yaml
```diff
@@ -0,0 +1,16 @@
+---
```
bruelea, line 1: This sample failed to claim a vlan when I tested it. "{\"site\":[\"Select a valid choice. DM-Akron is not one of the available choices.\"]}"

config/samples/netbox_v1_vlan.yaml
```diff
@@ -0,0 +1,16 @@
+---
```
bruelea, line 1: If two vlans with only a different metadata.name and vid are created, they "overwrite" each other in NetBox. Is the name field needed or could metadata.name be used? Instead of by name the vlans could be looked up by site and id.

config/samples/netbox_v1_vlanclaim_range.yaml
```diff
@@ -0,0 +1,16 @@
...
+    app.kubernetes.io/name: netbox-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: vlanclaim-range-sample
+spec:
+  tenant: "Dunder-Mifflin, Inc."
+  site: "DM-Akron"
+  description: "some description"
+  comments: "your comments"
+  preserveInNetbox: true
+  vidRangeStart: 200
```
bruelea: For this sample a vlan with id 100 was claimed, instead of an id between 200 and 299.

internal/controller/vlanclaim_controller.go
```diff
@@ -0,0 +1,326 @@
...
+		return nil, cancelLock, ctrl.Result{}, NewDomainError("%w", err)
+	}
+
+	if vlanModel != nil {
+		logger.V(4).Info(fmt.Sprintf("reassign reserved vlan vid from netbox: %d", vlanModel.Vid))
+		return &vlanModel.Vid, cancelLock, ctrl.Result{}, nil
+	}
+
+	// vlan cannot be restored from netbox
+	if o.Spec.Vid != 0 {
```
bruelea, line 314: The other controllers of NetBox Operattor do not support to define the desired value in the claim. This option could be a good addition to the claim, but in this case the claim controller should check if the object is still free in NetBox before assigning the Vid to a vlan CR.

bruelea (changes requested):
Thanks for taking over the implementation of the controllers for vlans, @flynn-nrg .
The structure of the controllers and test coverage looks good! Currently, some changes are needed for the claim/update logic of the vlans in NetBox to make sure different custom resources won't interfere with each other.

internal/controller/vlanclaim_controller.go
bruelea: What's your use case for choosing a vid manually in advance? Could it also be covered by claiming form a range?

pkg/netbox/api/vlan_claim.go
flynn-nrg: Fixed in https://github.com/netbox-community/netbox-operator/pull/618/changes/0d21c80e6bea656d6224e30505bc5caf12c5b194

config/samples/netbox_v1_vlanclaim_range.yaml
flynn-nrg: Fixed in https://github.com/netbox-community/netbox-operator/pull/618/changes/0d21c80e6bea656d6224e30505bc5caf12c5b194

internal/controller/vlanclaim_controller.go
flynn-nrg: Fixed in https://github.com/netbox-community/netbox-operator/pull/618/changes/0d21c80e6bea656d6224e30505bc5caf12c5b194

Our use case is when provisioning a new datacentre there are some static vlan allocations that are performed outside of NetBox. These are fed into the system prior to any dynamically VIDs that follow after. 

With this change these are treated as a range of 1.

config/samples/netbox_v1_vlan.yaml
flynn-nrg: Addressed in https://github.com/netbox-community/netbox-operator/pull/618/commits/3524ee2a52ffa5bc8f1af3bce9ea8a79ff1e8b5d

PR #626, opened 2026-09-01
reset selected parentprefix if it does not meet size criteria

bruelea:
There's a race condition which can cause a prefixclaim to get stuck because the available prefixes in the parentPrefix which was selected for do not meet the size criteria anymore at the time when a prefix is claimed from the parent prefix. This PR introduces a fix, by reselecting the parentPrefix in this case.


alexandernorth (approved):
lgtm

PR #627, opened 2026-09-03
request less resources for netbox installation

bruelea:
This PR reduces the resource request for the netbox installation, to avoid that pods don't get schedules in test environments with don't have much resources.

### DIFF
--- a/kind/deploy-netbox.sh
+++ b/kind/deploy-netbox.sh
@@ -149,6 +149,7 @@ if [ -n "$NETBOX_IMAGE_REGISTRY" ]; then
 fi
 
 # Install NetBox
+# --force-conflicts: override fields taken over by kubectl (server-side apply) on re-runs
 ${HELM} upgrade --install netbox ${NETBOX_HELM_CHART} \
   --namespace="${NAMESPACE}" \
   --create-namespace \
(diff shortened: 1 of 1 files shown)


kind/deploy-netbox.sh
```diff
@@ -149,6 +149,7 @@ if [ -n "$NETBOX_IMAGE_REGISTRY" ]; then
 fi
 
 # Install NetBox
+# --force-conflicts: override fields taken over by kubectl (server-side apply) on re-runs
```
MaIT-HA: why is this comment here when we don't use it in the command?

kind/deploy-netbox.sh
```diff
@@ -157,11 +158,24 @@ ${HELM} upgrade --install netbox ${NETBOX_HELM_CHART} \
...
-  --set resources.limits.cpu="2000m" \
-  --set resources.limits.memory="2Gi" \
+  --set resources.requests.cpu="125m" \
+  --set resources.requests.memory="128Mi" \
+  --set resources.limits.cpu="1000m" \
+  --set resources.limits.memory="1Gi" \
   --set redis.image.repository="bitnamilegacy/redis" \
+  --set redis.master.resourcesPreset="none" \
+  --set redis.master.resources.requests.cpu="50m" \
+  --set redis.master.resources.requests.memory="64Mi" \
```
MaIT-HA, line 167: I'm not too familiar with the setup, don't we also need .resources.limits?

kind/deploy-netbox.sh
```diff
@@ -157,11 +158,24 @@ ${HELM} upgrade --install netbox ${NETBOX_HELM_CHART} \
...
+  --set redis.replica.resourcesPreset="none" \
+  --set redis.replica.resources.requests.cpu="50m" \
+  --set redis.replica.resources.requests.memory="64Mi" \
+  --set valkey.auth.password="password" \
+  --set valkey.primary.resourcesPreset="none" \
+  --set valkey.primary.resources.requests.cpu="50m" \
+  --set valkey.primary.resources.requests.memory="64Mi" \
+  --set valkey.replica.resourcesPreset="none" \
+  --set valkey.replica.resources.requests.cpu="50m" \
+  --set valkey.replica.resources.requests.memory="64Mi" \
```
MaIT-HA, line 177: same question as for redis, don't we also need .resources.limits?

kind/deploy-netbox.sh
bruelea:
```suggestion
```

bruelea: missed this comment when I removed the flag

kind/deploy-netbox.sh
bruelea: Limits could be added, but to keep it simple I've only set requests here.

MaIT-HA: fine by me 👍

kind/deploy-netbox.sh
MaIT-HA: as discussed in the comment above

MaIT-HA (approved):

